PR #790, opened 2022-10-12
fix: Add alternate selector for subject container

wegry:
Keep `GmailThreadView.prototype.addNoticeBar` from throwing.

Screen Shot 2022-10-12 at 13 54 12


bkorobeinikov (approved):
lgtm

Issue #789, opened 2022-10-12
Bug: ThreadViewHandler not firing in Preview mode in gmail


teamstand-anup:
When the gmail is in preview mode (Select split vertically) and you click on any thread the threadViewHandler (function attached to registerThreadViewHandler) is not firing.
Happening in Chrome browser version Version 106.0.5249.119 (Official Build) (64-bit)
Though it doesnot seem to be browser version related.
The same thing worked last night so I am guessing that this is due to some changes in gmail dom maybe.

Thanks in advance

kuhnster01:
We are seeing this issue when an email is opened with the Gmail reading pane active. I think there may also be a related issue without the Gmail reading pane inactive.

aleemstreak:
Do you have repro steps for the related issue?

aleemstreak:
Also a minimal extension that reproduces the issue would be very helpful

kuhnster01:
It’s late on our side of the world and we will try help tomorrow if we have new information to share. We briefly tried to look in the GitHub repro but I don’t think we know our way around it yet.

As explained in the description on this issue, when a user opens an email while the reading pane active, we are no longer getting the call so that we can identify which thread the user has clicked on. 

We still see the Thread-F info in the DOM.

wegry:
Closed by #791.

teamstand-anup:
This issue is back. Maybe another build for Gmail UI.
Can we re-open this issue or should we create a new ticket?

teamstand-shahrukh:
The classname has been updated in some regions from `nH PeIF1d` to `nH a98 iY`.

1TheMuffinMan:
This issue seems to be back again. Both `sdk.Conversations.registerMessageViewHandler` and `sdk.Conversations.registerThreadViewHandler` aren't being triggered for a conversation if the preview pane is activated via the split view (horizontal or vertical)

PR #791, opened 2022-10-12
fix: add support for new thread selectors

bkorobeinikov:
add `.PeIF1d` selector to monitor thread changes in split mode

fix for https://github.com/InboxSDK/InboxSDK/issues/789


ben1lam8:
I tested manually and it works fine on my side with.
Did not test with previous DOM for compatibility.

GW 👌🏻

wegry (approved):

PR #799, opened 2022-10-21
Fix/thread container extraction

ben1lam8:
See #789 

I don't know what the logger exactly expects as version values, but I reused what has been done in previous fixes...


wegry:
Thanks for the PR.

That version value looks fine. The idea around version values is trying to keep track of selector versions over time.

Issue #801, opened 2022-10-21
ComposeView.getToRecipients() doesn't work within registerRequestModifier callback


bkorobeinikov:
**Describe the bug**
Calling `ComposeView.getToRecipients()` during `ComposeView.registerRequestModifier` callback returns empty result

**To Reproduce**

1. Start new compose
2. Register `registerRequestModifier` handler for composeView
4. add one or multiple to recipients
5. click "send" 
6. when `registerRequestModifier` is called, access `composeView.getToRecipients()`

**Expected behavior**
`ComposeView.getToRecipients()` should return recipients set in the compose `To` field 

**Add any other context about the problem here if anything**
Initial research lead to believe that during `registerRequestModifier` invocation, compose view element no longer has recipients chips, but instead has only span elements that are shown when To field is unfocused.

wegry:
See https://github.com/InboxSDK/InboxSDK/pull/951#issuecomment-1642172398, but I cannot reproduce this issue as reported with `examples/compose-button-tooltip`.

aleemstreak:
@bkorobeinikov have a minimal example?

PR #802, opened 2022-10-21
parse to,cc,bcc from res req and pass it to `registerRequestModifier`

bkorobeinikov:
during `registerRequestModifier` callback composeview might not contain recipients chips html to be retrieved via css selectors. 

Provide to,cc,bcc contacts in the  composeParams argument of a `registerRequestModifier` callback

Closes https://github.com/InboxSDK/InboxSDK/issues/801

usage
```javascript
composeView.registerRequestModifier(async cp => {

     const allRecipients = [...cp.to, ..cp.cc, ...cp.bcc];

});
```

The reasoning behind this PR:
After clicking "send" in compose window, gmail changes html so recipients list no longer accessible by selectors that work when compose window is closed. Instead of recipients chips, gmail leaves behind span based recipients elements that are shown when TO/CC/BCC fields are collapsed. Because compose window gets hidden when user clicks "send"  and required html potentially might not be available, it is better to provide needed fields within composeParams argument of `registerRequestModifier` callback.


src/platform-implementation-js/views/compose-view.ts
```diff
@@ -443,6 +443,9 @@ export default class ComposeView extends EventEmitter {
 
   public registerRequestModifier(
     modifier: (composeParams: {
+      to: Contact[] | null;
```
wegry, line 446: I think this is a similar case of Flow types not matching the TS.

https://github.com/InboxSDK/InboxSDK/pull/802/files#r1003607449

wegry (approved):

src/injected-js/gmail/constants.ts
```diff
@@ -1,7 +1,12 @@
+import { Contact } from '../../platform-implementation-js/driver-interfaces/compose-view-driver';
+
 export type ComposeRequestType = 'FIRST_DRAFT_SAVE' | 'DRAFT_SAVE' | 'SEND';
 
 export type ComposeRequest = {
   draftID: string;
+  to: Contact[] | null;
```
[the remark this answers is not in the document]
bkorobeinikov: changed flow types to be `Array<Contact> | null` https://github.com/InboxSDK/InboxSDK/pull/802/commits/367459cff8223662ce76d2b349872d9ed84e7bec

src/platform-implementation-js/views/compose-view.ts
bkorobeinikov: changed flow types to be same as ts types https://github.com/InboxSDK/InboxSDK/pull/802/commits/367459cff8223662ce76d2b349872d9ed84e7bec

bkorobeinikov:
> Any loose ends to tidy up here?

I need to add a couple of log messages to track legacy pathways to make sure they are no longer invoked, but other than that the PR is ready, but ill hold off merging it until after the holidays

bkorobeinikov:
NOTE: PR was marked as merged by github because https://github.com/InboxSDK/InboxSDK/pull/825 was merged which was a copy of this branch. The changes in this PR weren't merged to the main branch.

PR #813, opened 2022-12-14
fix: optional chain for all of gmonkey thread pulling

wegry:
Prevent `TypeError: window.gmonkey.v2.getCurrentThread is not a function.` flavor errors in the SDK.

### DIFF
--- a/src/injected-js/gmail/setup-gmonkey-handler.ts
+++ b/src/injected-js/gmail/setup-gmonkey-handler.ts
@@ -1,4 +1,6 @@
-/* @flow */
+declare global {
+  var gmonkey: any | undefined;
+}
 
 export default function setupGmonkeyHandler() {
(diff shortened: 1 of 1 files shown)


src/injected-js/gmail/setup-gmonkey-handler.ts
```diff
@@ -38,26 +35,20 @@ export default function setupGmonkeyHandler() {
           }
         }
       } else {
-        threadId =
-          window.gmonkey &&
-          window.gmonkey.v2 &&
-          window.gmonkey.v2.getCurrentThread().getThreadId();
+        threadId = window.gmonkey?.v2?.getCurrentThread?.()?.getThreadId();
```
wegry, line 38: This optional chaining addition should be the only runtime change in this PR.

Macil (approved):

PR #816, opened 2022-12-27
feat(labels): allow thread row label title customization

dmy7r0:
## Description

Allows passing an HTML along with text for thread row labels.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/widgets/gmail-label-view.js
+++ b/src/platform-implementation-js/dom-driver/gmail/widgets/gmail-label-view.js
@@ -73,7 +73,7 @@ class GmailLabelView {
     this._updateBackgroundColor(labelDescriptor.backgroundColor);
     this._updateForegroundColor(labelDescriptor.foregroundColor);
     this._updateIconBackgroundColor(labelDescriptor.iconBackgroundColor);
-    this._updateTitle(labelDescriptor.title);
+    this._updateTitle(labelDescriptor.title, labelDescriptor.titleHtml);
     this._updateTextMaxWidth(labelDescriptor.maxWidth);
 
     this._labelDescriptor = labelDescriptor;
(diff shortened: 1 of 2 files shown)


src/platform-implementation-js/dom-driver/gmail/widgets/gmail-label-view.js
```diff
@@ -84,7 +84,11 @@ class GmailLabelView {
       return;
     }
 
-    querySelector(this._element, '.av').style.maxWidth = maxWidth;
+    try {
+      querySelector(this._element, '.av').style.maxWidth = maxWidth;
+    } catch (e) {
+      // noop
+    }
```
Macil, line 91:
```suggestion
    this._element.querySelector('.av')?.style.maxWidth = maxWidth;
```

The querySelector function imported from querySelectorOrFail is specifically a function built to throw on failure as opposed to the standard querySelector method, so this code would be more natural to do with the standard querySelector method if we don't want that behavior.

Macil: Wait, this is still a js file using Flow, and I don't think that supports optional chaining, unlike our Typescript files.

Macil (approved):

src/platform-implementation-js/dom-driver/gmail/widgets/gmail-label-view.js
wegry: This file shows off how hard it is to try converting Flow to TS piecemeal over years too. It's depended on by `gmail-thread-row-view` which has 5 flow files depending on it.

wegry:
```suggestion
    const childEl = this._element.querySelector('.av')
    if (childEl) {
      childEl.style.maxWidth = maxWidth;
    }
```

dmy7r0: This looks great, thanks for the suggestions.

wegry (approved):

Issue #805, opened 2022-11-02
Bug: ComposeView Presending and Sent event order and threadIds inconsistent


aviatoali:
**Describe the bug**
Fairly often composeView's presending and sent event handlers are firing out of order, and with incorrect thread id. It's consistently reproducible but the number of emails you need to send to observe it may vary. You'll always trigger it at least once within every ten emails though.

The "sent" event will fire before the "presending" event and carry the threadId of the previous email. After presending no additional sent event with the correct threadId triggers. Usually requires you send emails somewhat closely back to back (5-20 seconds)

**To Reproduce**

1. Send an email
2. Observe presend and sent event fire order
3. Repeat

**Expected behavior**
Presend and sent events for a specific email fire in sequence (1. presend, 2. sent) and/or sent event's event.getThreadID() method returns the current email's threadId, rather than the previous email's threadId

**Screenshots**
https://screen.ac/geuyNWOO

**Add any other context about the problem here if anything**
[Min repro repo](https://github.com/aviatoali/InboxSDKBug), made from InboxSDK's hello world repo, so plug in your sdk app id, run the project and open console while sending emails. Will log the presend event, the sent event with threadId and the actual threadId of the new message when it gets added to threadRowView. Would recommend being in the "Sent" folder when sending these emails to get the threadId logged by threadRowView listener for comparison with the threadId logged by sent event handler

fjaeger:
This might be related to what @shff is reporting in https://github.com/InboxSDK/InboxSDK/issues/788

aleemstreak:
Thanks for the minimal addon that can reproduce the issue @aviatoali . We'll take a look at this in mid January, but happy to review any PRs in the meantime.

wegry:
@aviatoali you might try to reproduce the original issue with #824 now merged. If the issue persists, feel free to reopen this issue.

aviatoali:
Our initial testing is showing the issue resolve on our end. Thanks for the fix, much appreciated!

Issue #788, opened 2022-10-11
Bug: registerThreadRowViewHandler is not fired on several Compose sendings


auchynnikau:
**Description**
`registerThreadRowViewHandler` is not fired after quick send from several Compose after pasting message content.

This functionality is used for tracking emails using [PandaDoc email tracker extension](https://www.pandadoc.com/free-email-tracker-for-gmail/). The same problem was noticed and reproduced using another extension for tracking emails - [Streak](https://www.streak.com/). Also happens after quick send and content pasting.

Looks like an issue already [mentioned in Google Groups](https://groups.google.com/g/inboxsdk/c/AjIoOP6bJTk) might be related to this case and have similar reason but using `registerComposeViewHandler`.

**Steps to Reproduce**

1. Open Compose popup;
2. Fill all fields;
3. Open second Compose popup;
4. Paste content and quick send message;
5. Send first opened message;
6. Event is not fired on second sending.

**Expected behavior**
Event fired on each sending action.

**Screenshots**
<img width="259" alt="Screenshot 2022-10-12 at 01 00 50" src="https://user-images.githubusercontent.com/33329898/195206316-9ff4e239-258b-463e-891f-ce11e6b85b50.png">

shff:
I can reproduce this as well.

For the second compose view, the "sent" event will be triggered prematurely, when the Draft is saved, but with the wrong ID. It won't be triggered when the email is sent.

fjaeger:
Same here! We have this issue not being able to reliable fetch the final message-id for subsequently sent emails 😱

wegry:
Resolved by #824.

PR #824, opened 2023-01-13
Handle batched compose responses

wegry:
Potentially related to #788 and #805.

# Background

From @bkorobeinikov elsewhere:

> gmail can batch multiple responses for different compose views within one request. 
> and we just parse the first one in the array of different thread infos for different compose views

Instead of just handling the first message id, handle each and emit events associated with them.

# TODO

- [ ] Filter events emitted to respective compose views based on message ID ([comment](https://github.com/InboxSDK/InboxSDK/pull/824#discussion_r1070014909))


src/injected-js/gmail/sync-compose-processor-20220909.ts
```diff
@@ -14,11 +14,29 @@ export function parseComposeRequestBody_2022_09_09(request: Array<any>) {
...
+
+    const maybeBody =
+      parseUpdateDraftResponseBody(thread) ||
+      parseSendDraftResponseBody(thread);
+
+    if (maybeBody) {
+      return [maybeBody];
+    }
+
+    return [];
```
bkorobeinikov, line 38: from this code looks like first thread is being parsed always, but in this gist 
https://gist.github.com/bkorobeinikov/cc1eaf367bd0a5fa84b9beb98a1b42b8#file-gistfile1-txt

second item has correct value based on draftId in the request

src/injected-js/gmail/setup-gmail-interceptor.js
```diff
@@ -322,11 +322,13 @@ export function setupGmailInterceptorOnFrames(
...
             try {
-              const responseParsed = parseComposeResponseBody(
+              const responsesParsed = parseComposeResponseBody(
                 connection.originalResponseText
               );
 
-              if (responseParsed) {
+              console.log('responsesParsed', responsesParsed);
+
+              for (const responseParsed of responsesParsed) {
```
bkorobeinikov, line 331: since response can contain data for different draftIds, we should not iterate through array, because basically then triggerEvent will be fired for incorrect draftId that is in this context.

not saying it should be implemented like this, but essentially what we looking for is eaither to pass DraftId into `parseComposeResponseBody` and expect result to be relevant to passed DraftId, or filter result `const responseParsed = responsesParsed.filter(r => r.draftId === draftId)`

src/injected-js/gmail/sync-compose-processor-20220909.ts
wegry: Can you link to the line you think the draft id should have? The PR as of 85ac92b seems to pick up two message ids in this test https://github.com/InboxSDK/InboxSDK/pull/824/files#diff-383214f3ceee76012312efab301a5a77b3511352e93d08eff206f136045adb0cR64 with the gist linked.

We maybe dealing with trees of data in the batch response instead of just nested arrays.

src/injected-js/gmail/setup-gmail-interceptor.js
wegry: See https://github.com/InboxSDK/InboxSDK/pull/824/files#diff-f1974847c9b840296a92177ba6e677644336edbbf2656c52118fb96a9b51a6a1R331 for my stab at the latter approach.

test/data/2022-09-09-cvOnDraftSaveBatched_request.json
```diff
@@ -0,0 +1,108 @@
...
+                      1,
+                      "borys@streak.com",
+                      "Borys Korobeinikov",
+                      null,
+                      null,
+                      null,
+                      null,
+                      null,
+                      null,
+                      "borys@streak.com"
```
bkorobeinikov, line 31: replace my name and email with smth like in other files 

```
"test@gmail.com", "Test Test"
```

test/data/2022-09-09-cvOnDraftSaveBatched_request.json
```diff
@@ -0,0 +1,108 @@
+[
```
bkorobeinikov, line 1: could we keep name formats similar to ther test files, there is an index if the files represent similar req/resp


i would suggest going with `2022-09-09-cvOnDraftSave_2_request` for example, this will allow adding new versions of such files and keep the sorting in the explorer view in a way that all files within on group are together

bkorobeinikov (approved):
lgtm

test/data/2022-09-09-cvOnDraftSaveBatched_request.json
wegry: `test/data/2022-09-09-cvOnDraftSaveBatched_2_request.json` seems like a reasonable extension of this name. I couldn't tell by eyeballing #825 if we have other batched responses present.

Ideally, if JSON had comments in it, I'd put the information into the file instead of just the file name.

test/data/2022-09-09-cvOnDraftSaveBatched_response.json
```diff
@@ -0,0 +1,290 @@
...
+              [
+                "a 9",
+                null,
+                1673633584145,
+                "thread-a:r-175230870092109435",
+                [
+                  [
+                    "msg-a:r-173578382580458011",
+                    [1, "test@gmail.com", "Test A. Test"],
+                    [[1, "korobeinikov+test1@outlook.com", "Testy"]],
```
bkorobeinikov, line 62: also this email

src/injected-js/gmail/sync-compose-processor-20220909.ts
```diff
@@ -14,11 +14,29 @@ export function parseComposeRequestBody_2022_09_09(request: Array<any>) {
...
 }
 
-export function parseComposeResponseBody_2022_09_09(response: Array<any>) {
-  return (
-    parseUpdateDraftResponseBody(response) ||
-    parseSendDraftResponseBody(response)
-  );
+export function parseComposeResponseBody_2022_09_09(
+  response: Array<any>
+): any[] {
```
bkorobeinikov, line 19: could we change any[] return type, it breaks the typecheck

src/injected-js/gmail/sync-compose-processor-20220909.ts
bkorobeinikov: my mistake, there is a wrapper array around thread array, didn't notice it

PR #826, opened 2023-01-17
Fix: parse emoji in thread view subject

wegry:
Gmail encodes emoji in a way that allows it to keep emoji on brand.

```html
<img data-emoji="🔧" class="an1" alt="🔧" aria-label="🔧" src="https://fonts.gstatic.com/s/e/notoemoji/15.0/1f527/72.png" loading="lazy">
```

This PR extracts `data-emoji` attributes if present to allow them to be reflected in `ThreadView.prototype.getSubject`.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.js
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.js
@@ -474,18 +474,17 @@ class GmailThreadView {
 
     if (subjectElement.querySelector('img[data-emoji]')) {
       return Array.from(subjectElement.childNodes)
-        .flatMap((c) => {
+        .map((c) => {
           if (c instanceof HTMLElement && c.nodeName === 'IMG') {
             const maybeEmoji = c.getAttribute('data-emoji');
-            return maybeEmoji ? [maybeEmoji] : [];
+            return maybeEmoji;
           }
 
           if (c.nodeName === '#text') {
-            return [c.wholeText];
+            return (c: any).wholeText;
           }
-
-          return [];
         })
+        .filter((x) => x != null)
         .join('');
     }
 
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.js
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.js
@@ -470,9 +470,26 @@ class GmailThreadView {
     var subjectElement = this._element.querySelector('.ha h2');
     if (!subjectElement) {
       return '';
-    } else {
-      return subjectElement.textContent;
     }
+
+    if (subjectElement.querySelector('img[data-emoji]')) {
+      return Array.from(subjectElement.childNodes)
+        .flatMap((c) => {
+          if (c instanceof HTMLElement && c.nodeName === 'IMG') {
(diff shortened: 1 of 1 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.js
```diff
@@ -470,9 +470,25 @@ class GmailThreadView {
...
+
+    if (subjectElement.querySelector('img[data-emoji]')) {
+      return Array.from(subjectElement.childNodes)
+        .map((c) => {
+          if (c instanceof HTMLElement && c.nodeName === 'IMG') {
+            const maybeEmoji = c.getAttribute('data-emoji');
+            return maybeEmoji;
+          }
+
+          if (c.nodeName === '#text') {
```
bkorobeinikov, line 483: could it be that there are some other nodes in the subject besides img and plain text that should've been serialized into string?

wegry: Do you have an example you think that you think might be affected?

https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent has 

> textContent gets the content of all elements, including [<script>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) and [<style>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/style) elements. In contrast, innerText only shows "human-readable" elements.

bkorobeinikov: what if there is a span element inside? i don't have real example, but i don't think we can disproof it also

wegry: Does Gmail allow html in subjects? If you can create a scenario where a `<span>` is present, I'll gladly harden this to include that codepath.

As it is, the span is a bit of a devil's proof: https://en.wikipedia.org/wiki/Probatio_diabolica

Subject strings I tested, among others

- `<span>span</span>`
- `b 💸 ぼく`
- `Re: [InboxSDK/InboxSDK] Fix: parse emoji in thread view subject (PR #826)`

bkorobeinikov: but the previous behavior was covering all the cases by having `subjectElement.textContent` why not keep this with emoji parsing?

wegry: > the previous behavior was covering all the cases

It wasn't covering emoji (the way Gmail encodes them).

The PR description (if you missed it) has the Gmail encoding of an emoji we try to parse with the SDK:

```
<img data-emoji="🔧" class="an1" alt="🔧" aria-label="🔧" src="https://fonts.gstatic.com/s/e/notoemoji/15.0/1f527/72.png" loading="lazy">
```

The `img` _isn't_ included in text-content.

bkorobeinikov: yep i understand that, anyway, seems like you 100% sure it is not possible to create subject with anything else than img tag and plain text (including other/custom software that sends emails) and also taking into account that it is how gmail decided to render subject (adding img tag for example is a custom gmail work i believe) then ok with me

PR #827, opened 2023-01-18
fix: TypeError in logging inboxSDKinjectedError events

### LINKED ISSUE #819: Bug: TypeError: Cannot read properties of null (reading 'message')
**Describe the bug**

https://github.com/InboxSDK/InboxSDK/blob/main/src/platform-implementation-js/lib/logger.ts#L96 seems to throw if `event.detail` is undefined cluttering up error reporting that depends on `console.*` methods.

**Expected behavior**

If `event.detail` is not defined, handle it instead of throwing.

wegry:
Closes #819

### DIFF
--- a/src/platform-implementation-js/lib/logger.ts
+++ b/src/platform-implementation-js/lib/logger.ts
@@ -90,7 +90,12 @@ export default class Logger {
     })();
 
     if (this._isMaster && typeof document !== 'undefined') {
-      document.addEventListener('inboxSDKinjectedError', (event: any) => {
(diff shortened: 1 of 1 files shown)


Macil:
Our own code creates these event objects. We should make sure we're always creating them right.

wegry:
@Macil I DMed the link to error reporting in Slack.

Macil (approved):

Macil:
Thanks. Also I failed to appreciate at first that this PR still treats this case as an error, just a better labeled one.

PR #833, opened 2023-01-27
Chore: menu icon url and insert position 

dmy7r0:


src/platform-implementation-js/namespaces/app-menu.ts
```diff
@@ -3,11 +3,12 @@ import { AppMenuItemView } from '../views/app-menu-item-view';
 
 export type AppMenuItemDescriptor = {
   name: string;
-  iconElement: string;
+  iconUrl: string;
```
bkorobeinikov, line 6: just a note that we probably should change documentation too after these changes

bkorobeinikov: is this is a breaking change, once the inboxsdk version is deployed will this change influece existing code in production (ours and other people that use inboxsdk?)

Issue #829, opened 2023-01-23
Question: Which browser versions should InboxSDK support?


wegry:
**Question(s)**

Should InboxSDK only support the browsers Gmail supports? Gmail's browser support targets: https://support.google.com/mail/answer/6557?hl=en. InboxSDK seems to be targeting older browsers at the moment.

PR #836, opened 2023-02-03
feat: Lighter sdk build

wegry:
We currently support ECMAScript 2015. Safari 14 (2018) is the oldest browser supported after this change.

We can't go higher than Safari 14 with the current browserify bundler setup because of library AST limitations in the version of `acorn-walk` used by `babelify`. 

Related to #829.


babel.config.js
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable flowtype/require-valid-file-annotation */
+
+module.exports = (api) => ({
+  presets: [
+    [
+      '@babel/preset-env',
+      {
+        ...(api.env('test') ? { targets: { node: 'current' } } : {}),
+        corejs: 3.27,
+        useBuiltIns: 'entry',
```
Macil, line 10: Using `useBuiltIns: 'entry'` (and `corejs`) on babel/preset-env would make the InboxSDK include global polyfills. I don't think that would be a good idea for us, and it works against babel/plugin-transform-runtime's corejs option which includes local polyfills.

wegry: Dropped in [772ce6d](https://github.com/InboxSDK/InboxSDK/pull/836/commits/772ce6d2eb90e8ec47c40d8cd6cd933fe87821f8)

Longer term, I wonder what `transform-runtime` gets us as library (and not a bundle) should MV3 ever become the norm.

https://github.com/babel/babel/issues/10759#issuecomment-628130147 points out some issues around regenerator (which we won't need in Safari 14+ support).

Macil: babel/plugin-transform-runtime makes it so 1) instead of every JS-produced file including its own definition of helpers like _interopRequireDefault, it's imported from one place, 2) because we use the corejs option with babel/plugin-transform-runtime, it includes local polyfills for some newer JS features. 1 seems generally useful but 2 seems like something we can safely rip out as we have been using the old version of corejs (`"corejs": 2`) and I believe all browsers we target now have natively implemented everything we used to rely on corejs 2 for.

I'd be a fan of this PR cutting things down further by removing the `corejs: 3` option on babel/plugin-transform-runtime and adding `regenerator: false`.

yarn.lock
```diff
@@ -1467,7 +1467,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@babel/runtime-corejs2@npm:^7.0.0, @babel/runtime-corejs2@npm:^7.19.2":
+"@babel/runtime-corejs2@npm:^7.0.0":
```
Macil, line 1470: This dependency will go away entirely if you update the ud dependency.

Macil:
Looks good to me if you test the build with Streak successfully.

yarn.lock
wegry: [d4a8d8e](https://github.com/InboxSDK/InboxSDK/pull/836/commits/d4a8d8e9e19ce9ef2024e94d7e4e535496de43da) did this with a 

```
yarn up ud
yarn up -R ud
```

babel.config.js
wegry: [d4a8d8e](https://github.com/InboxSDK/InboxSDK/pull/836/commits/d4a8d8e9e19ce9ef2024e94d7e4e535496de43da) should have `corejs` removed once more.

gulpfile.babel.ts
```diff
@@ -189,6 +189,7 @@ async function browserifyTask(options: BrowserifyTaskOptions): Promise<void> {
   })
     .transform(
       babelify.configure({
+        presets: ['@babel/preset-env'],
```
Macil, line 192: Why is this specified here?

wegry: From https://github.com/babel/babelify#babelify-

> As of [Babel 6.0.0](http://babeljs.io/blog/2015/10/29/6.0.0/) there are no plugins included by default. For babelify to be useful, you must also include some [presets](http://babeljs.io/docs/plugins/#presets) and/or [plugins](http://babeljs.io/docs/plugins/#transform).

wegry: This was indeed a spurious bit of config. We chatted about it being unnecessary offline.

PR #843, opened 2023-02-16
Fix ComposeView.attachFiles() by avoiding variable shadowing

### LINKED ISSUE #842: Bug: error while calling the ComposeView.attachFiles() method in MV2 mode
Hello, I'm using your APIs in MV2 mode (with remote loading).
We've been unable to use the ComposeView.attachFiles() since a few days, the following error is returned:
```
platform-implementation.js:formatted:32877 Error logged: ReferenceError: Cannot access 'r' before initialization
    at Object._attachFiles (platform-implementation.js:formatted:40270:55)
    at Object.attachFiles (platform-implementation.js:formatted:40292:29)
    at p.attachFiles (platform-implementation.js:formatted:62371:67)
    at attachFiles (search.js:908:35)
    at Object.insertProposal (search.js:928:19) 
```
Indeed if you look at the relevanti minified code hosted on:
https://www.inboxsdk.com/build/platform-implementation.js
you can see the following (beautified):
```
            async _attachFiles(e, t) {
                this._hideDropzones();
                const n = (0,
                s.default)((()=>(0,
                S.simulateDragEnd)(this._element, e)));
                try {
                    let e = !0;
                    for (let r of a.default.partition(r, 3)) {
                 [...]
```
As you can see the minifier produce invalid code, at the 8th line it should have been ` for (let r of a.default.partition(e, 3)) {`

1. load inboxsdk in MV2 mode: `InboxSDK.load(2, 'YOUR_APP_ID_HERE').then(function(sdk){})`
2. register a composeView handler: `sdk.Compose.registerComposeViewHandler(function(composeView){})`
3. call the attachFiles method: `composeView.attachFiles()`

Alaanor:
Hello,

As suggested by @aleemstreak in https://github.com/InboxSDK/InboxSDK/issues/842#issuecomment-1430076502 , I made a PR to fix #842.

The problem was well described in the bug issue and the JS error is straightforward. The code was trying to access a variable before the initialization. I couldn't figure out _why_ this happened only now, but the issue was a name shadowing that should have worked but somehow in the minified version it confused the name.

The fix is simple, we just use another variable name to avoid the variable shadowing. I could reproduce the issue locally using the `main` branch and changing the name fixed it.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view.js
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view.js
@@ -1142,17 +1142,17 @@ class GmailComposeView {
     const endDrag = once(() => simulateDragEnd(this._element, files));
     try {
       let firstLoop = true;
-      for (let files of t.partition(files, 3)) {
+      for (let partitionedFiles of t.partition(files, 3)) {
         if (firstLoop) {
           firstLoop = false;
         } else {
           await delay(500);
         }
 
-        simulateDragOver(this._element, files);
+        simulateDragOver(this._element, partitionedFiles);
         await waitFor(() => this._dropzonesVisible(), 20 * 1000);
         const dropzone = this._findDropzoneForThisCompose(inline);
-        simulateDrop(dropzone, files);
+        simulateDrop(dropzone, partitionedFiles);
         endDrag();
         await waitFor(() => !this._dropzonesVisible(), 20 * 1000);
       }


wegry:
> I couldn't figure out why this happened only now.

@Alaanor this bug looks like it was introduced with #836. Switching from ES5 compatible output to Safari 14 removed babel iteration helpers. Switching from Safari 9 to 10 in the targets option of [this babel REPL](https://babeljs.io/repl/#?browsers=Safari%209&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABMOcAUAPAlIg3gKEWTgCdE0AbAUykQ0TmEQCMBDEzLHAooiBAM5xqAOgpwA5p0SFEAX3wL8oSLAQt2aAJ7dZiEjRAkkAbQDkYMyIMAHKqyjaRrLAF1F-ZajS5WALkQARjksIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Ctypescript&prettier=false&targets=&version=7.20.15&externalPlugins=&assumptions=%7B%7D) shows how.

PR #853, opened 2023-02-28
feat: loading-icon-support

wegry:
Add placeholder for loading into experimental collapsible panels.


src/platform-implementation-js/views/collapsible-panel-view.ts
```diff
@@ -38,25 +43,42 @@ export class CollapsiblePanelView extends (EventEmitter as new () => TypedEmitte
     HOVER: 'aJu',
     COLLAPSED_HOVER: 'bym',
   } as const;
-  #panelDescriptor: AppMenuItemPanelDescriptor | undefined;
-  #element?: HTMLElement;
+  panelDescriptor: AppMenuItemPanelDescriptor;
```
wegry, line 46: This should be an accessor so we can run `this.#update` against it.

dmy7r0 (approved):
Works great

PR #857, opened 2023-03-02
feat: update app menu icons on theme swap

wegry:
When swapping from light <-> dark themes within gmail, update icons in the app menu panel accordingly.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/gmail-driver/track-gmail-styles.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/gmail-driver/track-gmail-styles.ts
@@ -1,3 +1,4 @@
+import kefirBus from 'kefir-bus';
 import Logger from '../../../lib/logger';
 import waitFor from '../../../lib/wait-for';
 import GmailElementGetter from '../gmail-element-getter';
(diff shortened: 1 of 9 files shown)


src/platform-implementation-js/dom-driver/gmail/gmail-driver/track-gmail-styles.ts
```diff
@@ -84,28 +92,38 @@ export default function trackGmailStyles() {
...
+
+      stylesStream.emit({
+        type: 'theme',
+        isDarkMode: newDarkTheme,
+      });
     }
   }
 
-  waitFor(
-    () => document.querySelector('.TO .TN') && document.querySelector('.aio')
```
wegry, line 91: Swap back to this form of `waitFor`.

dmy7r0 (commented):
**UPD:** Fixed

When changing the theme I'm getting the error:
```
gmail-app-menu-item-view.ts:127 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'default')
    at GmailAppMenuItemView._updateIcon2 (gmail-app-menu-item-view.ts:127:72)
 ```

Screenshot 2023-03-02 at 3 23 27 PM

wegry:
@dmy7r0 do we want themes to be optional? They're currently required. I suspect you're calling the code from untyped JS.

PR #869, opened 2023-03-14
fix: getDensity method

Alex-Chopard:
Hi,
I've recently found that the class `inboxsdk__gmail_density_compact` was always set on the body even if the actual density setting was set to "Default" or "Comfortable".

I've made the changes in the method `getDensity` to detect if the density was set to Compact or Default/Comfortable.
Sadly, they are no visual/class changes between "Default" and "Comfortable" density!

I've used the key `default` for the default density, but it can be `cozy` to keep better retro compatibility, but `default` looks more relevant with the setting's name.

Steps to reproduce:
1. Load any extension using the InboxSDK
2. In the Gmail "Quick settings" switch the density from Compact to Default, the class `inboxsdk__gmail_density_compact` will always be on the body element

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/gmail-driver/track-gmail-styles.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/gmail-driver/track-gmail-styles.ts
@@ -5,25 +5,15 @@ import GmailElementGetter from '../gmail-element-getter';
 
 const RGB_REGEX = /^rgb\s*\(\s*(\d+),\s*(\d+),\s*(\d+)\s*\)/;
 
-function getDensity(): string {
-  // get the padding amount on the left-nav-menu entries to find the active
-  // density setting value.
-  const navItemOuter = document.querySelector('.TO .TN');
-  if (!navItemOuter) {
-    Logger.error(new Error('Failed to find nav item outer element'));
-    return 'compact';
+function getDensity(): 'compact' | 'default' {
+  const navItemElement = document.querySelector('.aim')
+  if (!navItemElement) {
+    Logger.error(new Error('Failed to find a nav item element'));
+    return 'default'
   }
-  const padding = parseInt(
-    getComputedStyle(navItemOuter).getPropertyValue('padding-top'),
-    10
-  );
-  if (padding >= 6) {
-    return 'comfortable';
-  } else if (padding >= 3) {
-    return 'cozy';
-  } else {
-    return 'compact';
-  }
+
+  const rect = navItemElement.getBoundingClientRect()
+  return rect?.height === 24 ? 'compact' : 'default'
 }
 
 const navItemSelector = '.aio' as const;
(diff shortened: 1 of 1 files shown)


Macil:
Thanks for this. While we've never documented these classnames as a public API, we at Streak do have some code outside of the InboxSDK relying on these specific classnames, so we're going to address that first before moving forward with this PR.

Just curious: are you making any use of these `inboxsdk__gmail_density_*` classnames in an extension using the InboxSDK?

Alex-Chopard:
Thank you for your response
I'm not using it yet, but I was looking for a way to change the style in a custom view when Gmail was in compact mode.

src/platform-implementation-js/dom-driver/gmail/gmail-driver/track-gmail-styles.ts
```diff
@@ -5,25 +5,15 @@ import GmailElementGetter from '../gmail-element-getter';
...
-  const padding = parseInt(
-    getComputedStyle(navItemOuter).getPropertyValue('padding-top'),
-    10
-  );
-  if (padding >= 6) {
-    return 'comfortable';
-  } else if (padding >= 3) {
-    return 'cozy';
-  } else {
-    return 'compact';
```
wegry, line 25: Because we at Streak use these class names, we'd need to keep this existing setup to merge this PR.

That said, @Alex-Chopard, you mentioned

> I was looking for a way to change the style in a custom view when Gmail was in compact mode.

There might be an alternative approach that meets both of our goals. I see a few potential options. In both, we'd keep the newer function.

1. Write new classnames with a different prefix. Instead of `'inboxsdk__gmail_density_' + currentDensity`, we'd have `'inboxsdk__apr2023_gmail_density_' + currentDensity
2. Have the newer function emit to a stream that we expose from the SDK. https://github.com/InboxSDK/InboxSDK/pull/869/files#diff-53e3b61a1eaf27bf59091d62566b209ec90bff97b2622ecb78e203c5392a6ab7R86 shows how we do so with light and dark modes. Maybe

wegry (changes requested):

src/platform-implementation-js/dom-driver/gmail/gmail-driver/track-gmail-styles.ts
wegry: An alternative @Macil brought up yesterday was taking our existing internal usages of `/inboxsdk__gmail_density__(cozy|comfortable|compact)/` and flattening down to use `compact` as our default. We'll get back to you on if that change goes through on our end.

wegry (approved):
Thanks for this, @Alex-Chopard. 

Because of 

> I've recently found that the class inboxsdk__gmail_density_compact was always set on the body even if the actual density setting was set to "Default" or "Comfortable".

we were able to determine that our internal usage of these classes wasn't used and now that we've removed them, using these two densities `'compact' | 'default'` makes sense.

Issue #873, opened 2023-03-15
Allow opting out of "Should Not Happen in non gmail websites" error


nirvparekh:
**Describe the bug**
When we open websites other than https://mail.google.com, it shows error in browser developer console like the screenshot below

**To Reproduce**

1. configure pageWorld: '@inboxsdk/core/pageWorld.js' in webpack.config.js
2. make a build and install the extension in chrome
3. Open website other than https://mail.google.com and open developer console

**Expected behavior**
It should not throw any such errors if extension is installed with this library.

**Screenshots**
![image](https://user-images.githubusercontent.com/40424777/225303289-34fcc9ac-33c6-46ab-b8f6-2022b48ea153.png)

wegry:
Which version of `@inboxsdk/core` are you seeing this error with?

wegry:
@nirvparekh mind creating a minimum, reproducible example? 

It's not clear to me if the extension in the reproduction steps is intended to be used both within `https://mail.google.com` and on others? If that's the case, we'd be willing to accept a PR.

nirvparekh:
> @nirvparekh mind creating a minimum, reproducible example?
> 
> It's not clear to me if the extension in the reproduction steps is intended to be used both within `https://mail.google.com` and on other domains? If that's the case, we'd be willing to accept a PR.

Thanks @wegry for showing interest in this issue.
Please check this repo: [https://github.com/nirvparekh/inbox-issue-reproduce](https://github.com/nirvparekh/inbox-issue-reproduce), I've created minimal app to reproduce this issue.

There are actually two issues:

1. when it is `https://mail.google.com` :

![image](https://user-images.githubusercontent.com/40424777/231671923-ccf24601-4a6b-4730-ba28-1a6ced29576e.png)

2. other than `https://mail.google.com` :

![image](https://user-images.githubusercontent.com/40424777/231672112-54256264-9075-42f4-a8c9-91f9968157f5.png)

wegry:
Thanks for the repro repo.

For 1:

You might create a separate issue. A quick glance at the minified code shows a `_headerObj`

```sh
$ rg -l -u  "_headersObj"
build/contentScript.js
node_modules/@inboxsdk/core/inboxsdk.js
node_modules/@inboxsdk/core/inboxsdk.js.map
```

This particular issue is present in both `@inboxsdk/core@1.0.0` and (the latest as of writing) `@inboxsdk/core@1.1.3`.

For 2:

[...]

PR #875, opened 2023-03-15
chore: drop vestigial inbox-driver switching by origin

wegry:


src/platform-implementation-js/platform-implementation.js
wegry, line 1: Is the `Driver` v. `GmailDriver` distinction something we want to keep around if we make a change like this?

src/platform-implementation-js/platform-implementation.js
```diff
@@ -240,28 +239,17 @@ export function makePlatformImplementation(
...
 
   const origin: string =
     (process.env.NODE_ENV === 'test' && global.__test_origin) ||
     document.location.origin;
-  const DriverClass = DRIVERS_BY_ORIGIN[origin];
-  if (!DriverClass) {
-    console.log('InboxSDK: Unsupported origin', origin);
-    // never resolve
-    return new Promise((resolve, reject) => {});
-  }
```
Macil, line 260: We need to keep the logic that if we're on an unrecognized domain we don't initialize the driver. It's common enough for people to import the InboxSDK in a content script that gets run on other domains. We could axe the console.log though: that was present to discourage people from remote-loading us into non-Gmail domains and giving us a large permission we didn't need, but we don't remote-load in the now recommended npm build so it doesn't matter much now.

src/platform-implementation-js/platform-implementation.js
Macil: Good question. Let's say no, so we're not keeping an unnecessary abstraction that might not pay off.

src/platform-implementation-js/platform-implementation.js
wegry: [b81ce52](https://github.com/InboxSDK/InboxSDK/pull/875/commits/b81ce52b738292abcda8e30f62783af49e68e199)

src/platform-implementation-js/platform-implementation.js
wegry: Dropped in [b81ce52](https://github.com/InboxSDK/InboxSDK/pull/875/commits/b81ce52b738292abcda8e30f62783af49e68e199).

src/platform-implementation-js/platform-implementation.js
wegry: > We need to keep the logic that if we're on an unrecognized domain we don't initialize the driver.

Keeping this logic has the added benefit of patching up #873.

PR #876, opened 2023-03-15
chore: remove stale comment

wegry:
Between fa70a731 and 0ab54548, we had a fallback get flipped from `true` to `false`. Update to rematch the comment with corresponding logic.


src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
```diff
@@ -225,7 +225,7 @@ const GmailElementGetter = {
         ?.getAttribute('aria-expanded') === 'true' ??
       // Default to true if we can't find the element because
       // the majority of users seem to favor that version.
-      false
+      true
```
dmy7r0, line 228: Defaulting to true I'm afraid is the action that will break things. I am happy to talk about it more offline.

PR #877, opened 2023-03-16
Feature Request: Add tooltip to NavItemViews

Alex-Chopard:
Add the same tooltip as Gmail use for his labels.

Screenshot 2023-03-16 at 12 01 38

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-nav-item-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-nav-item-view.ts
@@ -912,6 +912,10 @@ export default class GmailNavItemView {
     if (this._expandoElement) {
       this._expandoElement.title = `Expand ${name}`;
     }
+
+    this._element.firstElementChild?.setAttribute('data-tooltip', name);
+    this._element.firstElementChild?.setAttribute('data-tooltip-align', 'r');
+
     this._name = name;
   }
 
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-nav-item-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-nav-item-view.ts
@@ -66,6 +66,7 @@ export type NavItemDescriptor = {
     | DropdownButtonAccessoryDescriptor
     | null;
   type: keyof NavItemTypes;
+  tooltipAlignment: 'left' | 'top' | 'right' | 'bottom' | null;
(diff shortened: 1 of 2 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-nav-item-view.ts
```diff
@@ -912,6 +912,10 @@ export default class GmailNavItemView {
     if (this._expandoElement) {
       this._expandoElement.title = `Expand ${name}`;
     }
+
+    this._element.firstElementChild?.setAttribute('data-tooltip', name);
```
wegry, line 916: Could we add these attributes as [NavItemDescriptor](https://github.com/InboxSDK/InboxSDK/blob/c08fab3b9d85c1459633220a1fddf6bbc723e126/src/platform-implementation-js/dom-driver/gmail/views/gmail-nav-item-view.ts#L51) options so they're opt in?

With `data-tooltip-align`, for example, users may be in a rtl layout where they'd want alignment on the left.

Alex-Chopard: Totally agree, and with a value set to `null` it will allow users to not use this feature

src/platform-implementation-js/dom-driver/gmail/views/gmail-nav-item-view.ts
```diff
@@ -66,6 +66,7 @@ export type NavItemDescriptor = {
     | DropdownButtonAccessoryDescriptor
     | null;
   type: keyof NavItemTypes;
+  tooltipAlignment: 'left' | 'top' | 'right' | 'bottom' | null;
```
wegry, line 69: Using explicit `null` would require TS callers to update call sites.

```suggestion
  tooltipAlignment?: 'left' | 'top' | 'right' | 'bottom' | null;
```

Alex-Chopard: The property is within `Partial` type, so, it's already optional as all other properties (except the name), no changes will be required for existing users.
https://github.com/InboxSDK/InboxSDK/blob/02657eb133f1e7ed14612be63d3a2f16d5e03fa1/src/platform-implementation-js/dom-driver/gmail/views/gmail-nav-item-view.ts#L51-L70

wegry (approved):

PR #878, opened 2023-03-20
Fix custom view sizing

Macil:
This PR fixes the issues where custom views don't have the proper 16px margin at the bottom of the screen like native views, causing the page to be made too tall which causes HTMLElement.scrollIntoView() calls to scroll document.body itself, which isn't normally scrollable and causes content at the top of the page to be irreversibly scrolled out of view.


aleemstreak:
🙏

PR #879, opened 2023-03-23
Fix nav item view subtitle

Alex-Chopard:
- Add `subtitle` and `iconUrl` in `NavItemDescriptor` declaration type
- Replace the nav item view's subtitle instead of appending the new value upon update

I've also added my tests in the `custom-view` example, but they can be removed if you don't want to keep it


examples/custom-view/kefir.js
wegry, line 1: I'm not sure we'd want to commit the full minified `kefir` source here. Could we ignore this and just have it built into the correct place?

src/platform-implementation-js/dom-driver/gmail/views/gmail-nav-item-view.ts
```diff
@@ -963,7 +965,7 @@ export default class GmailNavItemView {
     }
   }
 
-  private _updateSubtitle(navItemDescriptor: any) {
+  private _updateSubtitle(navItemDescriptor: NavItemDescriptor) {
```
wegry, line 968: Seems sensible.

wegry (changes requested):
Thanks for helping remove some `any`s. This change seems reasonable, but it'd be nice if we could avoid adding a minified library to the repo if possible though.

src/platform-implementation-js/dom-driver/gmail/views/gmail-nav-item-view.ts
Alex-Chopard: `_updateSubtitle` has only one reference in `_updateValues` where the navItemDescriptor is already typed as a `NavItemDescriptor`, so, here, navItemDescriptor will always match the `NavItemDescriptor` type.
https://github.com/InboxSDK/InboxSDK/blob/c632cc01ab4a63e0e082a8cf1ac19398999d3d75/src/platform-implementation-js/dom-driver/gmail/views/gmail-nav-item-view.ts#L1026-L1039

PR #880, opened 2023-03-23
Feature Request: Emit event when NavItemView is inserted

Alex-Chopard:
Instead of using a `setTimeout` or a `MutationObserver`, it would be great to be notified by an event (`inserted` / `created`) when the NavItemView has been inserted in the DOM and when he is queryable with a `querySelector`.


aleemstreak:
What are you trying to accomplish with this? What  would you do after the
event is fired?

Alex-Chopard:
The purpose is to be able to query the `HTMLElement` of the `NavItemView` as soon as it has been inserted to modify it, use his x y coordinates to position a dialog, etc...

In my case, I need to point a dialog on a newly created NavItemView
Screenshot 2023-03-23 at 15 47 42

wegry (approved):

Macil (approved):

PR #881, opened 2023-03-28
chore: adopt CSS modules

wegry:
# What's in here?

## Added
- CSS Module support
- Browser support that matches the more modern "last two versions of Safari" we use at Streak that weren't possible with browserify. 

## Removed
- browserify from the build
- watching is done through webpack, which works with Node 18.12 and up unlike the unmaintained `watchify`.
- Our home rolled `IDMAP_` has been dropped in favor of CSS Module `:local`
- Because of the IDMAP_ removal, the old distinction between "shared" and "gmail" css has been removed as well.
- inbox.css is no longer hooked up

# TODO

- [x] Handling multiple SDK versions (see #885)
- [x] Maintain UMD output for npm build
- [x] Maintain UMD output for remote build
- [x] Source maps


gulpfile.babel.ts
```diff
@@ -6,30 +6,19 @@ const packageJson = JSON.parse(
...
 import exec from './src/build/exec';
-import babelify from 'babelify';
-import lazyPipe from 'lazypipe';
-import concat from 'gulp-concat';
-import addsrc from 'gulp-add-src';
 
 const sdkFilename = 'inboxsdk.js';
 
 const args = stdio.getopt({
+  // --watch is a node flag that doesn't seem to work after Node 18.12 with gulp.
```
wegry, line 20: Node 18.15 seems to recognize `--watch` and then fail. `-w` makes it through to the underlying gulp script though.

gulpfile.babel.ts
```diff
@@ -175,108 +125,131 @@ async function browserifyTask(options: BrowserifyTaskOptions): Promise<void> {
...
+      // https://github.com/webpack/changelog-v5/issues/10
+      new webpack.ProvidePlugin({
+        Buffer: ['buffer', 'Buffer'],
+        // https://stackoverflow.com/a/64553486 for default-encoding
+        // https://github.com/algolia/places/issues/847#issuecomment-748202652 For Algolia
+        process: 'process/browser',
       }),
-      { extensions: ['.js', '.tsx', '.ts'] }
-    )
-    .transform('redirectify', { global: true });
```
wegry, line 204: The `lodash/once` overwrite we were doing previously looked like it was shaving ~1KB off output size. There may have been a perf rationale for it that's been lost to time though. @Macil, if you remember, you might weigh in.

src/platform-implementation-js/lib/inject-script-EMBEDDED.ts
```diff
@@ -1,19 +1,17 @@
...
 
   const script = document.createElement('script');
   script.type = 'text/javascript';
 
-  const originalCode = fs.readFileSync(
-    __dirname + '/../../../dist/pageWorld.js',
-    'utf8'
-  );
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const originalCode = require('raw-loader!../../../dist/pageWorld.js');
```
wegry, line 11: Unfortunately, this version interestingly seems to get into a rebuild cycle sometimes. I have a feeling this affected the browserify version as well.

dmy7r0 (commented):
so far so good, looking forward to seeing this PR complete and merged  🙌

gulpfile.babel.ts
```diff
@@ -203,11 +203,14 @@ async function browserifyTask(options: BrowserifyTaskOptions): Promise<void> {
...
       ...(options.standalone
         ? {
-            libraryTarget: 'umd',
-            libraryExport: 'default',
-            library: options.standalone,
+            library: {
+              export: 'default',
+              name: options.standalone,
+              type: 'window',
+            },
```
Macil, line 210: What's the effect of this change?

wegry: I was thinking about this config for npm usages this morning, and this's probably not what we want for all four of the current build targets this PR touches (npm, remote, integratedPageWorld, and test).

`integratedPageWorld`, `remote`, and `test` (for now) builds should be using `library.type` `window` instead of UMD imo.

We could have UMD as the output format still for npm. Browserify was doing 

```
--standalone -s    Generate a UMD bundle for the supplied export name.
                   This bundle works with other module systems and sets the name
                   given as a window global if no module system is found.
```

That said, ESM has been de rigueur for output for some time. I'd prefer to output ESM (ES 2020) for our npm builds. 

For another library's perspective, `uuid` v9's changelog:

> Remove the minified UMD build from the package.
> 
> Minified code is hard to audit and since this is a widely used library it seems more appropriate nowadays to optimize for auditability than to ship a legacy module format that, at best, serves educational purposes nowadays.
> 
> For production browser use cases, users should be using a bundler. For educational purposes, today's online sandboxes like replit.com offer convenient ways to load npm modules, so the use case for UMD through repos like UNPKG or jsDelivr has largely vanished.

wegry: Maintaining the UMD version for `npm` output until we make the breaking change of ESM output, I think, is the way to go for now. Whether or not we want `UMD` for `remote` is an open question.

package.json
```diff
@@ -114,32 +97,27 @@
...
     "flow_check": "flow check",
     "lint": "eslint . --report-unused-disable-directives",
-    "lint-fix": "eslint . --ext js,jsx,ts,tsx,flow --fix",
+    "lint:fix": "eslint . --ext js,jsx,ts,tsx,flow --fix",
     "prettier-check": "prettier --check .",
     "prettier-fix": "prettier --write .",
     "puppeteer": "jest -c test/chrome/jest.config.js --runInBand",
     "start": "gulp default --watch --reloader",
-    "test": "yarn run lint && yarn run prettier-check && yarn run flow_check && tsc"
+    "test": "jest"
```
Macil, line 120: prettier-check step is missing now. Maybe make lint call it, or otherwise make sure ci calls it.

wegry: [0cc6f3f](https://github.com/InboxSDK/InboxSDK/pull/881/commits/0cc6f3fff30d86ecdc6dc79ae7df1b3cc82df7a3)

dmy7r0 (approved):

src/platform-implementation-js/lib/inject-script-EMBEDDED.ts
```diff
@@ -1,19 +1,16 @@
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const fs = require('fs');
+// const fs = require('fs');
```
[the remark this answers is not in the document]
wegry: [964bdf4](https://github.com/InboxSDK/InboxSDK/pull/881/commits/964bdf4d49c11c6926575b709739897673a608ab)

PR #885, opened 2023-03-29
fix: handle multiple sdk versions css

wegry:
At Streak, we've had issues where two versions of the SDK will race each other to insert the SDK's two style tags. Depending on the order styles are inserted, the later loading SDK using extension ends up in a bad state.

This PR follows up on #881 and uses webpack's style-loader to insert SDK style tags in the order in which they were published (after this change).

# TODO

- [x] Support loading an arbitrary number of style tags created by css module imports.


dmy7r0:
Went over this and let's try Algo Variant 1:

Inbox_SDK_algos(1)

wegry:
This PR'll need a few bits of rework before merging:

- CSS should be minified
- Each SDK version should only insert one style tag.
- Order doesn't matter with globals.

wegry:
Superseded by #895.

PR #891, opened 2023-04-04
experiment: algo variant 1

wegry:
Related to [#885](https://github.com/InboxSDK/InboxSDK/pull/885#issuecomment-1495044801).

The runtime certainly seems to work, but it looks like source maps are broken by the fact that each style tag has its styles overridden by the last file added. An example: if gmail.css is added first and collapsible-panel.module.css second, gmail.css erroneously has its source maps point at collapsible-panel.module.css.


wegry:
@Macil has pointed out that loading CSS globals in any order probably isn't solving too many problems. We'll keep the spirit `singletonStyleTag` insertion logic from this PR in our end solution though.

PR #895, opened 2023-04-10
fix: partially bring back older idMap impl

wegry:
Prevents issues where a @inbox/core@>=1.1.0 is unable to find the Gmail sidebar after an older @inbox/core using extension loads. This issue was introduced in #881.

From @Macil:

> ContentPanelViewDriver's constructor expects an element with the classname idMap('app_sidebar_waiting_platform') to exist already. The constructor is called from the addThreadSidebarContentPanel method of GmailAppSidebarView. GmailAppSidebarView's own constructor detects whether it's the first SDK instance, and if it is, constructs GmailAppSidebarPrimary which creates the element.
> 
> So it looks like it is the issue I brought up: if a different SDK loaded first, it will create the idMap('app_sidebar_waiting_platform') element, mark itself as the primary, and then other SDK instances will expect to find an existing idMap('app_sidebar_waiting_platform')  element.

This fix uses the older, rechristened `legacyIdMap` for selectors that are implicitly assumed to be stable across SDK versions. Styles are still associated with the newer CSS module build from #881.

### DIFF
--- a/src/common/version.ts
+++ b/src/common/version.ts
@@ -1,7 +1,9 @@
 // This is in its own file so that updates to the version value don't cause a
 // reload of everything.
 
-export const BUILD_VERSION: string = process.env.VERSION!;
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment -- SDK_VERSION is injected by webpack
+///@ts-ignore
+export const BUILD_VERSION: string = SDK_VERSION;
 
 if ((module as any).hot) {
   (module as any).hot.accept();
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-app-sidebar-view/primary/index.js
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-app-sidebar-view/primary/index.js
@@ -30,6 +30,11 @@ import fakeWindowResize from '../../../../../lib/fake-window-resize';
 import addCompanionThreadIconArea from './add-companion-thread-icon-area';
 import addCompanionGlobalIconArea from './add-companion-global-icon-area';
 import addToIconArea from './add-to-icon-area';
+import cx from 'classnames';
+import {
+  sidebarWaitingPlatformClassName,
+  sidebarWaitingPlatformSelector,
+} from '../../../../../driver-common/sidebar/constants';
 
 const ACTIVE_ADD_ON_ICON_SELECTOR = '.aT5-aOt-I-KO';
 const COMPANION_SIDEBAR_CONTENT_CLOSED_SHADOW_CLASS = 'brC-brG-btc';
@@ -671,11 +676,14 @@ class GmailAppSidebarPrimary {
 
     if (
       !((document.body: any): HTMLElement).querySelector(
-        '.' + idMap('app_sidebar_waiting_platform')
+        sidebarWaitingPlatformSelector
       )
     ) {
       const waitingPlatform = document.createElement('div');
-      waitingPlatform.className = idMap('app_sidebar_waiting_platform');
+      waitingPlatform.className = cx(
+        sidebarWaitingPlatformClassName,
+        idMap('app_sidebar_waiting_platform')
+      );
       ((document.body: any): HTMLElement).appendChild(waitingPlatform);
     }
 
--- a/src/platform-implementation-js/driver-common/sidebar/constants.ts
+++ b/src/platform-implementation-js/driver-common/sidebar/constants.ts
@@ -1,0 +1,34 @@
+import cx from 'classnames';
+import { legacyIdMap } from '../../lib/idMap';
+
+const cssGlobal = 'inboxsdk__app_sidebar_waiting_platform';
+
+const enum SidebarClassName {
+  /** v1.1.0 CSS module selector, */
+  v1_1_0 = 'inboxsdk__a0oshmeH6N_2vLtc3DNT',
+  /** v1.1.1 CSS module selector, */
+  v1_1_1 = 'inboxsdk__BV_FA4qVT4qYxs3LbEU4',
+}
+
+export const sidebarWaitingPlatformSelector = [
+  cssGlobal,
+  SidebarClassName.v1_1_0,
+  SidebarClassName.v1_1_1,
+  // Pre CSS module selector <1.1.0
+  legacyIdMap('app_sidebar_waiting_platform'),
+]
+  .map((x) => '.' + x)
+  .join(', ');
+
+/**
+ * The CSS local corresponding to this SDK's classname is intentionally in
+ * this classname. This is because this classname
+ * is used as a selector for disparate bits of code,
+ * and the CSS local is only used for styling.
+ */
+export const sidebarWaitingPlatformClassName = cx(
+  cssGlobal,
+  SidebarClassName.v1_1_0,
+  SidebarClassName.v1_1_1,
+  legacyIdMap('app_sidebar_waiting_platform')
+);
--- a/src/platform-implementation-js/lib/idMap.ts
+++ b/src/platform-implementation-js/lib/idMap.ts
@@ -1,4 +1,46 @@
 import * as s from '../style/gmail.css';
+import Sha256 from 'sha.js/sha256';
+
+const cache: Map<string, string> = new Map();
+let seed: string | null = null;
+
+/**
+ * Pass a name and you'll get a unique identifier associated with that name.
+ * The returned identifier will only contain alphabetic characters.
+ *
+ * @deprecated a holdover from the old idMap runtime implementation. Use `idMap` instead unless you specifically need the same classname as old versions of the InboxSDK used.
+ */
+export function legacyIdMap(name: string): string {
+  const id = cache.get(name);
+  if (id != null) return id;
+
+  if (seed == null) {
+    seed = document.documentElement.getAttribute('data-map-id');
+    if (seed == null) {
+      // Make the seed change every hour: the seed is based on a hash of the
+      // current timestamp in hour resolution.
+      const hasher = new Sha256();
+      hasher.update(
+        'PWVe' + Math.floor(Date.now() / (1000 * 60 * 60)) + 'PYiE0'
+      );
+      seed =
+        hasher.digest('hex').slice(0, 16) +
+        (process.env.NODE_ENV === 'development' ? 'x' : '');
+      document.documentElement.setAttribute('data-map-id', seed);
+    }
+  }
+
+  const hasher = new Sha256();
+  hasher.update('4iYi29W' + name + ':' + seed + 'jn2mPvTG');
+  const newId = hasher
+    .digest('hex')
+    .slice(0, 16)
+    .replace(/[0-9]/g, (match) =>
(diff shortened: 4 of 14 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-app-sidebar-view/primary/index.js
```diff
@@ -166,7 +167,7 @@ class GmailAppSidebarPrimary {
       document.createElement('div'));
     threadSidebarContainerEl.classList.add(
       'thread_app_sidebar',
-      idMap('app_sidebar_container'),
+      ...[legacyIdMap, idMap].map((f) => f('app_sidebar_container')),
```
Macil, line 170: We never look for an element with the classname legacyIdMap('app_sidebar_container') produced by another SDK (or the current one), so we don't need legacyIdMap for 'app_sidebar_container'.

wegry: [5624339](https://github.com/InboxSDK/InboxSDK/pull/895/commits/5624339b7568f255ab9a0b14402cfd3f25e8722c)

src/common/version.ts
```diff
@@ -1,7 +1,11 @@
 // This is in its own file so that updates to the version value don't cause a
 // reload of everything.
 
-export const BUILD_VERSION: string = process.env.VERSION!;
+declare global {
+  var SDK_VERSION: string;
+}
```
Macil, line 6: When we generate .d.ts files, will this end up in there? We don't want to expose this to consumers.

src/platform-implementation-js/lib/idMap.ts
```diff
@@ -1,4 +1,46 @@
...
+import Sha256 from 'sha.js/sha256';
+
+const cache: Map<string, string> = new Map();
+let seed: string | null = null;
+
+/**
+ * Pass a name and you'll get a unique identifier associated with that name.
+ * The returned identifier will only contain alphabetic characters.
+ *
+ * @deprecated a holdover from the old idMap runtime implementation. Use `idMap` instead.
```
Macil, line 11:
```suggestion
 * @deprecated a holdover from the old idMap runtime implementation. Use `idMap` instead unless you specifically need the same classname as old versions of the InboxSDK used.
```

src/common/version.ts
wegry: Our generated `.d.ts` files aren't currently hooked to anything (at least in our internal usage). Point taken though. This would indeed show up in `@inboxsdk/core/` and pollute consuming projects...

src/platform-implementation-js/driver-common/sidebar/constants.ts
```diff
@@ -0,0 +1,27 @@
...
+import { legacyIdMap } from '../../lib/idMap';
+
+const cssGlobal = 'inboxsdk__app_sidebar_waiting_platform';
+
+const enum SidebarClassName {
+  /** v1.1.0 CSS module selector, */
+  v1_1_0 = 'inboxsdk__a0oshmeH6N_2vLtc3DNT',
+  /** v1.1.1 CSS module selector, */
+  v1_1_1 = 'inboxsdk__BV_FA4qVT4qYxs3LbEU4',
+}
```
Macil, line 11: Weird to use an enum for this since "SidebarClassName" isn't really a type, and const enums have weird compatibility catches with Babel. Could this just be an object instead?

src/platform-implementation-js/driver-common/sidebar/constants.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const sidebarWaitingPlatformSelector = `.${cssGlobal},
+  ${[SidebarClassName.v1_1_0, SidebarClassName.v1_1_1]
+    .map((x) => '.' + x)
+    .join(', ')},
+  .${
+    // Pre CSS module selector <1.1.0
+    legacyIdMap('app_sidebar_waiting_platform')
+  }` as const;
```
Macil, line 20: It's weird how this declaration uses a mix of template strings and array mapping to create its result. Here's a more consistent version:

```suggestion
export const sidebarWaitingPlatformSelector = [
  cssGlobal,
  SidebarClassName.v1_1_0,
  SidebarClassName.v1_1_1,
  // Pre CSS module selector <1.1.0
  legacyIdMap('app_sidebar_waiting_platform'),
]
    .map((x) => '.' + x)
    .join(', ');

File: src/platform-implementation-js/driver-common/sidebar/constants.ts (hunk 4 above)
Comment (wegry) [reply src/platform-implementation-js/driver-common/sidebar/constants.ts:11]: > const enums have weird compatibility catches with Babel

Not true since the end of 2021.

https://github.com/babel/babel/pull/13324

An enum a shorthand to group the two values together and it compiles to an object at build time.

File: src/platform-implementation-js/driver-common/sidebar/constants.ts [hunk 6]
```diff
@@ -0,0 +1,27 @@
...
+
+export const sidebarWaitingPlatformClassName = cx(
+  cssGlobal,
+  SidebarClassName.v1_1_0,
+  SidebarClassName.v1_1_1,
+  legacyIdMap('app_sidebar_waiting_platform')
+);
```
Macil, line 28: We need to add the current idMap value here so that our CSS applies to the element:
```suggestion
export const sidebarWaitingPlatformClassName = cx(
  cssGlobal,
  SidebarClassName.v1_1_0,
  SidebarClassName.v1_1_1,
  legacyIdMap('app_sidebar_waiting_platform'),
  idMap('app_sidebar_waiting_platform')
);
```

Macil: oh, I see where you set the classname you're adding `idMap('app_sidebar_waiting_platform')` there. I think it's a little confusing it's not here instead but don't feel strongly.

wegry: I have it separately to highlight that it's not related to this compatibility classname/selector. I could add a comment to that effect.

src/platform-implementation-js/dom-driver/gmail/views/gmail-app-sidebar-view/primary/index.js
```diff
@@ -662,20 +669,25 @@ class GmailAppSidebarPrimary {
       this._companionSidebarOuterWrapper.previousElementSibling;
     if (contentContainer) {
       contentContainer.classList.add(
-        idMap('companion_container_app_sidebar_in_use')
+        ...[legacyIdMap, idMap].map((f) =>
+          f('companion_container_app_sidebar_in_use')
+        )
```
Macil, line 674: No need for legacyIdMap on "companion_container_app_sidebar_in_use".

src/common/version.ts
wegry: [5f7c6a6](https://github.com/InboxSDK/InboxSDK/pull/895/commits/5f7c6a6317dd22c455b315f8aeb84f79b1946731)

src/platform-implementation-js/dom-driver/gmail/views/gmail-app-sidebar-view/primary/index.js
```diff
@@ -194,9 +199,9 @@ class GmailAppSidebarPrimary {
           threadSidebarContainerEl.getBoundingClientRect();
         const titleBars = Array.from(
           threadSidebarContainerEl.querySelectorAll(
-            `.${idMap('app_sidebar_content_panel')}.${idMap(
+            `.${legacyIdMap('app_sidebar_content_panel')}.${legacyIdMap(
```
Macil, line 202: This is broken because we create this element with the classnames using only idMap.

But also this is an element we create ourselves so we should only use idMap here anyway, not legacyIdMap, for "app_sidebar_content_panel", "expanded", "app_sidebar_content_panel_top_line".

src/platform-implementation-js/dom-driver/gmail/views/gmail-app-sidebar-view/primary/index.js
```diff
@@ -367,7 +372,9 @@ class GmailAppSidebarPrimary {
         buttonContainer = existingButtonContainer;
       } else {
         buttonContainer = document.createElement('div');
-        buttonContainer.className = idMap('sidebar_button_container');
+        buttonContainer.className = cx(
+          [legacyIdMap, idMap].map((f) => f('sidebar_button_container'))
+        );
```
Macil, line 377: only need idMap for "sidebar_button_container"

src/platform-implementation-js/driver-common/sidebar/constants.ts
Macil: a comment would be good

src/platform-implementation-js/dom-driver/gmail/views/gmail-app-sidebar-view/primary/index.js
wegry: 1fd12d80

src/platform-implementation-js/dom-driver/gmail/views/gmail-app-sidebar-view/primary/index.js
```diff
@@ -666,16 +673,19 @@ class GmailAppSidebarPrimary {
       );
     }
     this._companionSidebarContentContainerEl.classList.add(
-      idMap('app_sidebar_in_use')
+      ...[legacyIdMap, idMap].map((f) => f('app_sidebar_in_use'))
```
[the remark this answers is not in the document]
wegry: 432e9809

src/platform-implementation-js/driver-common/sidebar/constants.ts
Macil: Sure, but both the `const` and `enum` parts do nothing here.

src/platform-implementation-js/dom-driver/gmail/views/gmail-app-sidebar-view/primary/index.js
wegry: e38cc028

src/platform-implementation-js/driver-common/sidebar/constants.ts
wegry: [64ec92d](https://github.com/InboxSDK/InboxSDK/pull/895/commits/64ec92d4fb9da777d71cf908a8674f103d660c27)

src/platform-implementation-js/driver-common/sidebar/constants.ts
wegry: With [01238b2](https://github.com/InboxSDK/InboxSDK/pull/895/commits/01238b2e453b7e76fd8a2e8fee18cf892691b2fe), which I had in #885, the `const enum` is a zero cost abstraction and inlined at runtime.

Macil: I think it's still weird because a "SidebarClassName" type makes no sense and doesn't get us anything more than an object would here, but this is non-blocking.

Macil (approved):

PR #896, opened 2023-04-11
chore: minify inlined CSS

wegry:
Mopping up from #895. Make sure to minify inlined CSS with csso.


dmy7r0 (commented):
> csso

[Canadian Society of Surgical Oncology](https://www.csso.surgery/)?

Jokes aside, looks good but didn't test it. Would be helpful to see the before/after of the built file and size to validate the modification method.

Macil (approved):

wegry:
@dmy7r0 For size changes, bundlephobia doesn't have great granularity, but it's something like 5/6ths of the 1.1.2 size. https://bundlephobia.com/package/@inboxsdk/core@1.1.3

In case bundlephobia is down:
Screenshot 2023-04-12 at 07 38 27

PR #898, opened 2023-04-12
feat: stylelint

wegry:
More lingering bits from #895. Introduce stylelint here that we use in internal projects.

The CSS most affected might be in `gmail.css` formerly `shared.css` and `gmail.css` prior to #881. There are a number of duplicate selectors from when the two files were merged together to share the same CSS locals.

Other notable rules are the dropping of many stale vendor prefixes in the `examples/*` folder.


wegry:
Closing in favor of #912.

PR #900, opened 2023-04-26
Add replaceSubject to thread row view

jfmaggie:
Updated example extension `Thread Rows`

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.js
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.js
@@ -995,14 +995,22 @@ class GmailThreadRowView {
   }
 
   getSubject(): string {
-    if (this._elements.length > 1) {
-      return querySelector(this._elements[1], 'div.xS div.xT div.y6 > span[id]')
-        .textContent;
-    } else {
-      return querySelector(
-        this._elements[0],
-        'td.a4W div.xS div.xT div.y6 > span[id]'
-      ).textContent;
+    return this.#getSubjectSelector().textContent;
+  }
+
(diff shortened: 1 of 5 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.js
```diff
@@ -995,14 +995,22 @@ class GmailThreadRowView {
...
   getSubject(): string {
-    if (this._elements.length > 1) {
-      return querySelector(this._elements[1], 'div.xS div.xT div.y6 > span[id]')
-        .textContent;
-    } else {
-      return querySelector(
-        this._elements[0],
-        'td.a4W div.xS div.xT div.y6 > span[id]'
-      ).textContent;
+    return this._getSubjectSelector().textContent;
```
wegry, line 998: Do we want human readable elements only? If so

```suggestion
    return this._getSubjectSelector().innerText;
```

See https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext

wegry (approved):

src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.js
jfmaggie: I’m not sure the intended design of original getSubject. The documentation says `Returns a String of the subject of this thread.` it didn’t really say getting human readable content. Text content gives all the elements. I’d avoid changing the original API in this PR, it’s supposed to be adding a new method with no impact on existing usage of other methods. It’s more of an API change if any usage of getSubject() needs the all the elements string.

Macil (approved):

PR #901, opened 2023-05-02
fix: add 'bugfixes: true' for Safari 15 babel class fields codegen

wegry:
babel may be generating standards compliant code around class fields, but Safari 15 might have a bug that causes `ReferenceErrors`.

See https://github.com/babel/babel/issues/14289#issuecomment-1374547406 for how class fields could break in Safari 15 codegen.


wegry:
Superseded by #906.

PR #904, opened 2023-05-04
rollback browserlist to Safari v14

bkorobeinikov:
Safari 15 cannot handle this syntax when code is minified

```
import once from 'lodash/once';

class Hello {
  world = once(() => {
    return "hello world";
  });
}
```

the file that fails to be executed in Safari v15 is `src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.js`

the bug was made in the https://github.com/InboxSDK/InboxSDK/pull/881

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,2 +1,2 @@
 Chrome 100
-Safari 15
+Safari 14


wegry:
Did you try against https://github.com/InboxSDK/InboxSDK/pull/901?

wegry:
The issue this is alleviating looks like it's around terser generating incorrect code around class fields with closures assigned to them for Safari 15. It could also just be a Safari 15 bug as well.

```
import {once} from 'lodash';

class Foo {
  classFieldMethod = once(() => console.log())  
}
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields

wegry (approved):
Rolling back'll give us some breathing room for a fix.

We do have CSS that depends on Safari 15 in newer code, so in practice, this change doesn't mean that we actually support Safari 14 again.

PR #906, opened 2023-05-08
fix: reapply Safari 15 min

wegry:
Fix for #904. Our usage of CJS inside of ESM exports seems to generate a ReferenceError with our stack in Safari 15 when coupled with a class that's being assigned to `module.exports` within a ud defn function.

Swapping our browserslist default creates quite a bit of noise in our error logging, as it changes stack trace locations and increases stack size. ~It also increases our output code size by roughly 10%.~ _edit: bundlephobia doesn't rounds the two output sizes together, but we're in the ~1% range minified._

~Longer term, I'd like to remove the remainder of ud in its entirety as it doesn't work with TS and is thus blocking moving off flow.~ _See #910._

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,2 +1,2 @@
 Chrome 100
-Safari 14
+Safari 15
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,32 +1,50 @@
 /* eslint-disable flowtype/require-valid-file-annotation */
 
-module.exports = (api) => ({
-  presets: [
-    [
-      '@babel/preset-env',
+function isBabelRegister(caller) {
+  return !!(caller && caller.name === '@babel/register');
+}
+
+module.exports = (api) => {
+  const isRegister = api.caller(isBabelRegister);
+
+  return {
+    presets: [
+      [
+        '@babel/preset-env',
+        {
+          ...(!isRegister &&
+            !api.env('test') && {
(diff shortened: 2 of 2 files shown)


dmy7r0 (commented):
I was able to test this against Safari 15 and it is not breaking anymore.

Sidenote: It took me a decent amount of time to set up the env for testing and I wish it was simpler/automated in repetitive tasks.

Macil:
I'm very worried that this ud call isn't the only thing that could trigger this issue. If we switch to target Safari 15, then future code changes could hit this issue again.

I found an issue about the underlying Safari bug: https://github.com/babel/babel/issues/14289. It looks like the recommended workaround is to manually enable the `@babel/plugin-proposal-class-properties` babel plugin.

wegry:
> I found an issue about the underlying Safari bug: https://github.com/babel/babel/issues/14289. It looks like the recommended workaround is to manually enable the @babel/plugin-proposal-class-properties babel plugin.

@Macil see [d297787](https://github.com/InboxSDK/InboxSDK/pull/906/commits/d2977876c2f4b62da110e77255550a5b873a46c1).

Macil (approved):

wegry:
@Macil https://github.com/terser/terser/blob/master/CHANGELOG.md#v5176 may hint at the heart of the issue that underlies this patch. It could also explain why we hadn't necessarily seen the issue manifest in our downstream projects.

PR #910, opened 2023-05-25
chore: drop ud from default exports of classes

wegry:
Eases porting class based code to TS.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/gmail-driver/getSyncThreadFromSyncThreadId.js
+++ b/src/platform-implementation-js/dom-driver/gmail/gmail-driver/getSyncThreadFromSyncThreadId.js
@@ -1,6 +1,5 @@
 /* @flow */
 
-import { defn } from 'ud';
 import gmailAjax from '../../../driver-common/gmailAjax';
 import { extractThreadsFromThreadResponse } from '../gmail-sync-response-processor';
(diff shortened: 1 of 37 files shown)


Macil:
I posted about a solution for #906 in that thread that's not ud-related, so we shouldn't treat this issue as related to it now.

I'm not a big fan of completely purging ud from the codebase. The InboxSDK doesn't have hot-reloading support right now since the move to Webpack but I would like for us to get that back at some point. My comments about ud before about it being removable were mainly meant about if it gets in the way anywhere, like how in Typescript it doesn't properly pass through the types of ud-wrapped classes. I would be interested in this PR if it only removed ud from being used with classes, but kept it in other places like around functions and page-parser-tree options.

wegry:
> I posted about a solution for https://github.com/InboxSDK/InboxSDK/pull/906 in that thread that's not ud-related, so we shouldn't treat this issue as related to it now.

@Macil Description updated

> I would be interested in this PR if it only removed ud from being used with classes, but kept it in other places like around functions and page-parser-tree options.

[8efc33c](https://github.com/InboxSDK/InboxSDK/pull/910/commits/8efc33c19f3fab3b93fd219c14825f37286b4054) does just that. For simplicity's sake, only `/defn\([A-Z]\w+)/` ud usages are removed.

Macil (approved):

PR #912, opened 2023-05-30
chore: add stylelint while punting on duplicate selector merging

wegry:
Alternate and more conservative version of #898. This version runs `yarn stylelint --fix` to remove vendor-prefixes and use newer color formatting. It _doesn't_ try to merge duplicate selectors together. `csso`, our CSS minifier (post #896), is presumably already doing that.


src/platform-implementation-js/style/gmail.css
```diff
@@ -842,13 +829,13 @@ input.inboxsdk__x_close_button {
...
 .inboxsdk__gmail_action:active {
-  box-shadow: inset 0 1px 2px rgba(0, 0, 0, 0.1);
+  box-shadow: inset 0 1px 2px rgb(0 0 0 / 10%);
 }
 
 .inboxsdk__gmail_action:hover {
-  box-shadow: 0 1px 1px rgba(0, 0, 0, 0.05);
+  box-shadow: 0 1px 1px rgb(0 0 0 / 5%);
   background-color: #ededed;
-  background-image: -webkit-linear-gradient(top, #ededed, #eaeaea);
```
wegry, line 851: From chatting with @dmy7r0 offline (https://caniuse.com/?search=linear-gradient):

> Partial support in Safari and Older Firefox versions refers to not using premultiplied colors which results in unexpected behavior when using the transparent keyword as advised by the [spec](https://www.w3.org/TR/2012/CR-css3-images-20120417/#color-stop-syntax).

Because we support Safari 15 for now, we can just let the existing change be with a stylelint disable to get this through.

src/platform-implementation-js/style/gmail.css
```diff
@@ -1240,7 +1233,6 @@ isn't visible behind it. */
   border: none;
   display: inline-flex;
   justify-content: center;
-  outline: none;
```
wegry, line 1243: Why was this removed?

wegry: @dmy7r0 It looks like `outline` in this ruleset was just doubled up. The subsequent [one is here](https://github.com/InboxSDK/InboxSDK/pull/912/files#diff-4deb24efbdddeae857792fd63a7c2937123ae6c6f6863bd2bfbe6ccacdeaad88R1247)

src/platform-implementation-js/style/gmail.css
wegry: Reverted in f48defd4fab71f8ad38da3f6b0166d56c615994b

dmy7r0 (approved):

PR #913, opened 2023-06-02
chore: move a few more files to TS

wegry:
Moves some more files over to TS from Flow (syntactically), `platform-implementation` and `compose-view` being the biggest two. I suggest ignoring whitespace changes. To reduce noise, _there should be no runtime changes_. If you see any, please flag them.

Most ESLint rules have been disabled where they conflict for newly moved Flow to TS files for the sake of noise. ~One exception is `prefer-const` autofixed in a few spots.~ _Moved to #916_. A subsequent pass to remove file wide `@typescript-eslint/no-unused-variables` disabling might flesh out a good chunk of dead code in the form of unused imports. 

Other bits:

- `Kefir.later(n)` has been rendered as `Kefir.later(n, undefined)` to match the TS Kefir types.
- `flow check` has been removed
- ~`.js.flow` files have been removed. This is roughly 1/3 of the files touched in the PR.~ _Moved to #917_
- most `getEventStream` methods on views tend to be loosely typed and should be tightened over time.

PR #918, opened 2023-06-07
chore: drop unused variables left over from #913

wegry:
- remove ESLint overrides on `no-unused-var`
- drop gmail-thread-sidebar-view that doesn't seem to be used
- remove casts on `document.{body,head}`


src/injected-js/xhr-helper.ts
```diff
@@ -13,72 +12,37 @@ export default function xhrHelper() {
     // It's important to use fetch when possible because it's needed for
     // getDownloadURL() in Gmail v2: Gmail v2's ServiceWorker in Chrome causes
     // xhr.responseURL to have the wrong value (possibly a Chrome bug).
-    // TODO: This condition will always return true since this function is always defined. Did you mean to call it instead? ts(2774)
-    if (global.fetch as any) {
```
wegry, line 17: `fetch` is supported in every browser we support since #881. The corresponding `else` block for non-fetch support has been removed.

src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-sidebar-view.tsx
wegry, line 1: This file was referenced in src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts, but it looks like it was only used as a type and nothing more. Removing it means that src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-sidebar-view/add-icon-area.ts is unused as well.

src/platform-implementation-js/platform-implementation.ts
```diff
@@ -230,19 +229,6 @@ https://www.inboxsdk.com/docs/#RequiredSetup
     logger.eventSdkPassive('instantiate');
     const pi = new PlatformImplementation(driver, appId, opts);
 
-    // TODO: drop. This condition isn't possible.
-    if (origin === ('https://inbox.google.com' as any) && !opts.inboxBeta) {
```
wegry, line 234: #887 removed most other references to old `'https://inbox.google.com'`. Without the `any`, TS correctly notes that this block can never be hit because of a check further up in the function.

PR #932, opened 2023-06-22
docs: officially release AppMenu

wegry:
We've had the AppMenu in use at Streak for 3+ months now. We believe that the AppMenu is stabilized. This PR tweaks a bit of TSDoc to reflect that.

### DIFF
--- a/src/platform-implementation-js/views/app-menu-item-view.ts
+++ b/src/platform-implementation-js/views/app-menu-item-view.ts
@@ -25,7 +25,7 @@ import defer from '../../common/defer';
 
 type MessageEvents = {
   /**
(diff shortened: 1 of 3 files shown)


src/platform-implementation-js/views/app-menu-item-view.ts
```diff
@@ -24,8 +24,17 @@ import GmailElementGetter from '../dom-driver/gmail/gmail-element-getter';
 import defer from '../../common/defer';
 
 type MessageEvents = {
+  /**
+   * Fires when this AppMenuItemView's menuItem has a `leave` event triggered.
```
wegry, line 28:
```suggestion
   * Fires when this AppMenuItemView's menuItem has a `mouseleave` event triggered.
```

PR #934, opened 2023-06-27
chore: turn off minification for npm published builds by default

wegry:
If you're not using a bundler, using `inboxsdk.min.js` instead of `inboxsdk.js` should mirror the old behavior.

Related to debugging #933.

Without #938, note that with this change the minified `inboxsdk.min.js` is using terser@5.18.2, while the remote loaded SDK is using terser@5.16.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@ yarn-debug.log
 /.inboxsdk_test_secret
 yarn-error.log
 /packages/core/inboxsdk.js*
-packages/core/inboxsdk.min.js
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -14,6 +14,7 @@ yarn-debug.log
 /.inboxsdk_test_secret
 yarn-error.log
 /packages/core/inboxsdk.js*
+packages/core/inboxsdk.min.js
```
wegry, line 17:
```suggestion
/packages/core/inboxsdk.min.js
```

PR #936, opened 2023-06-27
feat: add collapseToggled event to AppMenu

wegry:
Allows timing events that require measuring around the width transition in collapsible panels.

### DIFF
--- a/src/platform-implementation-js/namespaces/app-menu.ts
+++ b/src/platform-implementation-js/namespaces/app-menu.ts
@@ -111,7 +111,7 @@ export default class AppMenu {
       open: boolean;
       /**
        * 'start' is when a `transactionstart` event is fired on a collapsible panel.
-       * 'end' is 200ms after 'start' and based on the transition time of width animation in Gmail. */
+       * 'end' is when a `transactioncancel` event is fired on a collapsible panel. */
       stage: 'start' | 'end';
     },
     unknown
@@ -129,33 +129,34 @@ export default class AppMenu {
         return;
       }
 
-      appMenu.parentElement!.addEventListener(
-        'transitionstart',
-        async ({ target }) => {
+      const onTransition =
+        (stage: 'start' | 'end') =>
+        async ({ target }: TransitionEvent) => {
+          const { TOGGLE_OPEN_STATE } = CollapsiblePanelView.elementCss;
+
           if (
             target instanceof HTMLElement &&
-            // Is the target a collapsible panel with the transition class?
             target.matches(CollapsiblePanelView.elementSelectors.NATIVE) &&
-            target.classList.contains(
-              CollapsiblePanelView.elementCss.TOGGLE_OPEN_STATE
-            )
+            ((stage === 'start' &&
+              target.classList.contains(TOGGLE_OPEN_STATE)) ||
+              // TOGGLE_OPEN_STATE is removed on transitioncancel TransitionEvents
+              stage === 'end')
           ) {
             emitter.emit({
               type: 'collapseToggled',
               open: this.isMenuCollapsed(),
-              stage: 'start',
-            });
-
-            // `transitionend` TransitionEvents don't seem to fire reliably for collapsible panels. This may
-            // be because .aak, the class applying transition properties, is removed before the transitionend would fire.
-            await new Promise((resolve) => setTimeout(resolve, 200));
-            emitter.emit({
-              type: 'collapseToggled',
-              open: this.isMenuCollapsed(),
-              stage: 'end',
+              stage,
             });
           }
-        }
+        };
+
+      appMenu.parentElement!.addEventListener(
+        'transitionstart',
+        onTransition('start')
+      );
+      appMenu.parentElement!.addEventListener(
+        'transitioncancel',
+        onTransition('end')
       );
     };
 
--- a/src/platform-implementation-js/namespaces/app-menu.ts
+++ b/src/platform-implementation-js/namespaces/app-menu.ts
@@ -1,7 +1,9 @@
+import { stream } from 'kefir';
 import { RouteView } from '../../inboxsdk';
 import GmailDriver from '../dom-driver/gmail/gmail-driver';
 import GmailElementGetter from '../dom-driver/gmail/gmail-element-getter';
 import { AppMenuItemView } from '../views/app-menu-item-view';
+import { CollapsiblePanelView } from '../views/collapsible-panel-view';
 
 type ThemedIcon =
   | string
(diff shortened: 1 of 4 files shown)


src/platform-implementation-js/namespaces/app-menu.ts
```diff
@@ -102,11 +104,84 @@ Items with CollapsiblePanels can also have accessories which provide primary act
...
+              CollapsiblePanelView.elementCss.TOGGLE_OPEN_STATE
+            )
+          ) {
+            emitter.emit({
+              type: 'collapseToggled',
+              open: this.isMenuCollapsed(),
+              stage: 'start',
+            });
+
+            // `transitionend` TransitionEvents don't seem to fire reliably for collapsible panels. This may
```
bkorobeinikov, line 149: I believe the `transitioncancel` will be called in this case

wegry: That does indeed seem to do the trick. See 16a0293a.

src/platform-implementation-js/namespaces/app-menu.ts
```diff
@@ -102,11 +104,85 @@ Items with CollapsiblePanels can also have accessories which provide primary act
...
+   * @note If the AppMenu is not shown, this stream will not emit any events.
+   */
+  get events() {
+    return this.#events;
+  }
+
+  /**
+   * @returns whether or not the AppMenu Burger is collapsed or not.
+   */
+  isMenuCollapsed() {
```
dmy7r0, line 182: Q: Shouldn't this be named "open" based on `isAppBurgerMenuOpen`?

src/platform-implementation-js/namespaces/app-menu.ts
```diff
@@ -102,11 +104,85 @@ Items with CollapsiblePanels can also have accessories which provide primary act
...
   #driver;
+  #events = stream<
+    {
+      type: 'collapseToggled';
+      /** Whether or not the AppMenuBurger is open or not. */
+      open: boolean;
+      /**
+       * 'start' is when a `transactionstart` event is fired on a collapsible panel.
+       * 'end' is when a `transactioncancel` event is fired on a collapsible panel. */
+      stage: 'start' | 'end';
```
bkorobeinikov, line 115: could you give more context why the 'stage' is necessary?

couple of things:
1. the event named `collapseToggle` but the property is `open`, lets make it either `openChanged` event or `collapseToggled` with `collapsed` property
2. What is the purpose of `stage` prop? I would imagine the consumer code would do smth like 
```
const isOpened = (event.open == false && stage === 'start' || event.open == true && stage === 'end')
```
if that is the case, lets remove `stage` prop from this event, and do this calculation within AppMenu to avoid such logic being duplicated in all places where it is needed.
3. the `transitionend` event should still be listened too, otherwise if for some reason the transition will actually complete (like gmail will fix a bug or something), the `transitioncancel` will never be called.

src/platform-implementation-js/namespaces/app-menu.ts
```diff
@@ -102,11 +104,85 @@ Items with CollapsiblePanels can also have accessories which provide primary act
...
+        };
+
+      appMenu.parentElement!.addEventListener(
+        'transitionstart',
+        onTransition('start')
+      );
+      appMenu.parentElement!.addEventListener(
+        'transitioncancel',
+        onTransition('end')
+      );
```
dmy7r0, line 160: I think `transitioncancel` should be handled as well.

src/platform-implementation-js/namespaces/app-menu.ts
```diff
@@ -102,11 +104,85 @@ Items with CollapsiblePanels can also have accessories which provide primary act
...
+            target.matches(CollapsiblePanelView.elementSelectors.NATIVE) &&
+            ((stage === 'start' &&
+              target.classList.contains(TOGGLE_OPEN_STATE)) ||
+              // TOGGLE_OPEN_STATE is removed on transitioncancel TransitionEvents
+              stage === 'end')
+          ) {
+            emitter.emit({
+              type: 'collapseToggled',
+              open: this.isMenuCollapsed(),
+              stage,
```
dmy7r0, line 148: The `open` value might get stale when this event gets triggered and handled. Personally, I'd emit just an event type and leave the value retrieval to the end user (via app menu API)

src/platform-implementation-js/namespaces/app-menu.ts
wegry: `transitioncanend` or `transitionend`?

src/platform-implementation-js/namespaces/app-menu.ts
wegry: > emit just an event type

Do you mean a `TransitionEvent`?

src/platform-implementation-js/namespaces/app-menu.ts
wegry: I'm ok with either. I feel like we've gone back and for on the name for collapsed or open since #849.

src/platform-implementation-js/namespaces/app-menu.ts
wegry: At least in the use case we have at Streak, I'd worry we'd end up double firing if something changed within Gmail around how transitions triggered if we handled both `transitioncancel` and `transitionend`.

src/platform-implementation-js/namespaces/app-menu.ts
wegry: Changed to `isMenuOpen`.

src/platform-implementation-js/namespaces/app-menu.ts
wegry: [079ff95](https://github.com/InboxSDK/InboxSDK/pull/936/commits/079ff957cebe68b578391319dd4e2ab8a3d74dc6) handles the three events and adds a caveat that `transitionend` doesn't currently seem to fire in TSDoc.

src/platform-implementation-js/namespaces/app-menu.ts
wegry: [079ff95](https://github.com/InboxSDK/InboxSDK/pull/936/commits/079ff957cebe68b578391319dd4e2ab8a3d74dc6) makes checking `AppMenu.prototype.isMenuOpen` the subscriber's responsibility.

src/platform-implementation-js/namespaces/app-menu.ts
wegry: a few responses:

> the event named collapseToggle but the property is open, lets make it either openChanged event or collapseToggled with collapsed property

1. @dmy7r0 and I have gone back and forth on `collapsed` v. `open`. One thing that bit us during `AppMenu` development was stale values. https://github.com/InboxSDK/InboxSDK/pull/936#discussion_r1244346685 reminded me that the subscriber checking toggleCollapsed is probably a better idea.

> What is the purpose of stage prop?

2. At least in the use case we're trying to solve in the repo downstream, which you might've missed [the link for above this thread](https://github.com/StreakYC/MailFoo/pull/14017), we _only_ immediately care about `transitionend`. Riffing on [this comment](https://github.com/InboxSDK/InboxSDK/pull/936#discussion_r1244346685) more, just emitting the `TransitionEvent` and having the subscriber do what they want with `TransitionEvent.type` being `transition{start,end,cancel}` seems fairly straightforward.

> the transitionend event should still be listened too, otherwise if for some reason the transition will actually complete (like gmail will fix a bug or something), the transitioncancel will never be called.

3. Added in [079ff95](https://github.com/InboxSDK/InboxSDK/pull/936/commits/079ff957cebe68b578391319dd4e2ab8a3d74dc6) with the caveat that as of writing, it is not called. It looks like Gmail is removing the class with a transition time of 200ms something like 190ms into the animation right now.

dmy7r0 (approved):
Thanks for the updates

PR #937, opened 2023-06-27
chore(nav item view): element getter

dmy7r0:


wegry (approved):
The types from this should just flow through on publish. I think https://github.com/InboxSDK/InboxSDK/blob/16a0293a46542e4cc54f78171436d94135dc59ab/src/inboxsdk.d.ts#L348 is re-exported here, but you might check with

`yarn link -r ~/InboxSDK/packages/core` and running tsc in an upstream project.

src/platform-implementation-js/views/nav-item-view.ts
```diff
@@ -140,6 +140,12 @@ export default class NavItemView extends EventEmitter {
       }
     );
   }
+
+  getElement() {
```
wegry, line 144: We might add TSDoc here. Something like:

```suggestion
  /**
   * Asynchronously get the DOM element containing the NavItemView
   */ 
  getElement() {
```

PR #940, opened 2023-07-03
experiment: switch sourceURL to sourceMappingURL for remote loaded SDK

wegry:


Macil:
There should already be a source mapping URL comment inside this injected script, or if not that's an issue with the build config and not something we should do here.
SourceURL is used intentionally here so dev tools can have a filename for this code instead of just "inline script", and so that the source mapping URL comment already in the injected script is resolved relative to this URL.

wegry:
> There should already be a source mapping URL comment inside this injected script.

@Macil There almost certainly is. [A private example from a project using the remote loaded SDK](https://streak.sentry.io/issues/4231269411/?environment=production&project=4504322969763840&query=is%3Aunresolved+plan%3Apaid&referrer=issue-stream&stream_index=6). 
Screenshot 2023-07-05 at 07 34 04. The source map seems to be off by a number of lines though.

PR #946, opened 2023-07-13
fix: remote build source maps

wegry:
This change should only affect remote and `integratedPageWorld` builds--not npm package output.

We've been having issues where our many of our SDK source maps appear to be incorrectly mapped to `gmail-attachment-card`. This may be because our remote builds feature inlining of `pageWorld.js` into `platform-implementation`.

This PR changes remote build code in the following ways:

- inlining source maps for pageWorld
- minification is re-enabled for remote builds
- Only `inboxsdk.js` is bundled as UMD. `platform-implementation` is now the webpack default executable
- Remove deprecated `raw-loader` in favor of webpack's asset modules

### DIFF
--- a/gulpfile.babel.ts
+++ b/gulpfile.babel.ts
@@ -132,8 +132,8 @@ const enum OutputLibraryType {
   ESM = 'module',
 }
 
-interface BrowserifyTaskOptions {
-  entry: string;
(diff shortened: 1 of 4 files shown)


wegry:
As this is, I don't think it'll help with our evaling of JS in remote builds...

PR #948, opened 2023-07-13
chore: simplify gulp tasks

wegry:
Two of our build targets, remote and integratedPageWorld, require a two separate webpack passes. This is because these two targets platform-implementation.js depends on inlining the pageWorld webpack entry point's output. This PR tries to highlight that difference between these two targets and our npm build. The npm build has been moved to an webpack.Configuration['entry'] type to reduce the distance between our repurposed browserify wrapper and webpack.

Salvaged from #946

### DIFF
--- a/gulpfile.babel.ts
+++ b/gulpfile.babel.ts
@@ -12,7 +12,7 @@ import extReloader from './live/extReloader';
 import fg from 'fast-glob';
 import exec from './src/build/exec';
 
-const sdkFilename = 'inboxsdk.js';
+const sdkFilename = 'inboxsdk';
 
 const args = stdio.getopt({
   // --watch is a node flag that doesn't seem to work after Node 18.12 with gulp.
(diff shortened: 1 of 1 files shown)


gulpfile.babel.ts
```diff
@@ -341,84 +318,106 @@ gulp.task('types', async () => {
...
+      },
+      'platform-implementation':
+        './src/platform-implementation-js/main-INTEGRATED-PAGEWORLD',
+    },
+    plugins: [
+      new webpack.SourceMapDevToolPlugin({
+        fileContext: '../',
+        // Adds hash to the end of the sourcemap URL so that
+        // remote code source maps are not cached erroneously by Sentry.
+        filename: '[file].map?hash=[contenthash]',
```
wegry, line 347: After speaking with @Macil, these comments may be being stripped out by our pre-OSS SDK version that we have currently deployed.

gulpfile.babel.ts
```diff
@@ -341,84 +318,106 @@ gulp.task('types', async () => {
...
-      return webpackTask({
-        devtool: 'remote',
-        entry: './src/platform-implementation-js/main-INTEGRATED-PAGEWORLD',
-        destName: 'platform-implementation.js',
-        afterBuild: setupExamples,
-      });
-    })
-  );
-  gulp.task('default', gulp.series('sdk', 'remote'));
+  config = {
```
wegry, line 327: We should inline the block from 405 invoking the webpack task both here and for the integratedPageWorld branch.

gulpfile.babel.ts
```diff
@@ -341,84 +318,106 @@ gulp.task('types', async () => {
...
+        {
+          encoding: 'utf8',
+        }
+      );
+    },
+  };
 }
 
+gulp.task('sdk', async () => {
+  if (args.remote || args.integratedPageWorld) {
```
wegry, line 406:
```suggestion
  // Only the npm build contains pageWorld as of writing.
  if (!config.entry.pageWorld) {
```

gulpfile.babel.ts
wegry: @Macil mentioned that https://github.com/InboxSDK/InboxSDK/pull/948#discussion_r1289181171 would assuage his concerns here.

gulpfile.babel.ts
```diff
@@ -351,84 +331,107 @@ gulp.task('types', async () => {
...
-  gulp.task('default', gulp.parallel('sdk', 'types'));
+  config = {
+    ...remoteCompatConfig,
+    devtool: 'inline-source-map',
+    entry: {
+      [sdkFilename]: {
+        library: {
+          export: 'default',
+          name: 'InboxSDK',
+          type: OutputLibraryType.Var,
```
wegry, line 376: We should use UMD for the integratedPageWorld to mirror the remote loaded version.

```suggestion
          type: OutputLibraryType.UMD,
```

gulpfile.babel.ts
```diff
@@ -351,84 +331,107 @@ gulp.task('types', async () => {
...
+    afterBuild: async () => {
+      setupExamples();
+      const { minify } = await import('terser');
+
+      const sourceOutput = await fs.promises.readFile(
+        'packages/core/inboxsdk.js',
+        'utf8'
+      );
+
+      const minified = await minify(sourceOutput);
```
wegry, line 406: In the interest of reducing build target support, we may remove this output.

We could always bring it back if a user needs it.

PR #951, opened 2023-07-19
chore: ComposeView implements ComposeViewDriver

### LINKED ISSUE #801: ComposeView.getToRecipients() doesn't work within registerRequestModifier callback

wegry:
Work around looking into #801.

Includes:

- Type work to see how `ComposeView` and its corresponding driver interact
- Types in examples

### DIFF
--- a/examples/app-menu/content.js
+++ b/examples/app-menu/content.js
@@ -1,4 +1,4 @@
-/// <reference path="types.d.ts" />
+/// <reference path="../types.d.ts" />
 
 function log() {
   console.log(...['app-menu'].concat(arguments));
--- a/examples/compose-tooltip-button/content.js
+++ b/examples/compose-tooltip-button/content.js
@@ -1,3 +1,5 @@
+/// <reference path="../types.d.ts" />
+
 InboxSDK.load(2, 'simple-example').then(function(inboxSDK) {
 	inboxSDK.Compose.registerComposeViewHandler(function(composeView){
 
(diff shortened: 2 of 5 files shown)


wegry:
I cannot reproduce #801 with this patch 

```diff
diff --git a/examples/compose-tooltip-button/content.js b/examples/compose-tooltip-button/content.js
index 7c80b0cc..a0849c74 100644
--- a/examples/compose-tooltip-button/content.js
+++ b/examples/compose-tooltip-button/content.js
@@ -2,6 +2,11 @@
 
 InboxSDK.load(2, 'simple-example').then(function(inboxSDK) {
 	inboxSDK.Compose.registerComposeViewHandler(function(composeView){
+    composeView.registerRequestModifier(function (req) {
+      console.log('req', req);
+      console.log('to recipients', composeView.getToRecipients());
+      return req;
+    });
 
 		window.composeView = composeView;
```

I've tried single and multiple users, but I suspect the error outlined in #801 is probably Streak specific code.

Issue #943, opened 2023-07-11
Bug: 


pbosakov:
**Describe the bug**
When switching to a different category tab in Gmail, errors are reported in the console:
```
pageWorld.js:2 Error in injected script Error: Errors in thread row parsing
    at P (pageWorld.js:2:1)
    at pageWorld.js:2:1
    at HTMLDocument.<anonymous> (pageWorld.js:2:1) {errors: [
        "failed to find timeSpan",
        "failed to find peopleDiv"
    ]}
```

**To Reproduce**

1. [Enable category tabs](https://support.google.com/mail/answer/3094499) in Gmail
2. Open the developer console
3. Open inbox and click to a different tab (e.g. _Promotions_)

**Screenshots**
![inboxsdk-screenshot](https://github.com/InboxSDK/InboxSDK/assets/1148312/1583be78-288f-4d41-94b6-011fa5918137)
![inboxsdk-console](https://github.com/InboxSDK/InboxSDK/assets/1148312/5afb4e6a-3be3-4f36-85ff-487e098b180d)

wegry:
Notably, it seems I can repro this issue on a Gmail account but not on a Google Workspace account.

wegry:
Digging a bit more: unfortunately, there is no `timeSpan` equivalent in the _Promotions_ or _Social _ tabs.

See https://github.com/InboxSDK/InboxSDK/blob/2dadf97d932691251f5dc508213819a95cc2358e/src/injected-js/gmail/thread-identifier/thread-row-parser.ts#L36 for the `timeSpan` element selector that works in the Primary tab.

`peopleDiv` _could_ be potentially be remapped to the left most column name. In the [example above](https://user-images.githubusercontent.com/1148312/252630560-1583be78-288f-4d41-94b6-011fa5918137.png) that'd be `Tripadvisor`.

Whether or not we could look up non Primary tab thread information based on the equivalent (or lack of one) is an open question.

wegry:
With respect to

> Whether or not we could look up non Primary tab thread information based on the equivalent (or lack of one) is an open question.

It appears that even with this [example extension](https://github.com/InboxSDK/InboxSDK/blob/main/examples/thread-rows/content.js) and this diff: 

```
diff --git a/src/injected-js/gmail/thread-identifier/thread-row-parser.ts b/src/injected-js/gmail/thread-identifier/thread-row-parser.ts
index 420be1f4..d7814feb 100644
[...]

PR #967, opened 2023-08-03
types: link ThreadRow interface with its impl

wegry:
- Removes `isSelected` method type that was present in `ThreadRowView` interface but not in the implementation.
- Notes method deprecations with TSDoc in a way the TS/JS Language Servers can pick up

Solidifies types before #943.


src/platform-implementation-js/views/thread-row-view.ts
```diff
@@ -1,35 +1,120 @@
...
   destroyed: boolean;
 
   constructor(threadRowViewDriver: GmailThreadRowView) {
     super();
     const members = {
       threadRowViewDriver,
     };
-    membersMap.set(this, members);
     this.destroyed = false;
+    membersMap.set(this, members);
```
wegry, line 94:
```suggestion
    membersMap.set(this, members);
    this.destroyed = false;
```

src/inboxsdk.d.ts
```diff
@@ -263,59 +264,10 @@ export interface User {
...
-  ): void;
-  getElement(): HTMLElement;
-  getVisibleDraftCount(): number;
-  getThreadIDIfStable(): string | null;
-  getThreadIDIfStableAsync(): Promise<string | null>;
-  addButton(threadRowButtonDescriptor: any): void;
-  getThreadID(): string;
-  getThreadIDAsync(): Promise<string>;
-  getDraftID(): Promise<string>;
-  isSelected(): boolean;
```
wegry, line 301: Note: this method doesn't seem to exist in the implementation.

PR #970, opened 2023-08-03
Rip loadScript and eval out of the npm (non-remote) build

Macil:
This PR rips out eval usage and the `InboxSDK.loadScript()` function from the npm (non-remote) version of the InboxSDK.
The `InboxSDK.loadScript()` function and its eval usage will still be present in the legacy remote version of the InboxSDK.

This is being done because eval and `InboxSDK.loadScript()` are unusable in Chrome MV3 extensions, and it's possible that their presence in the JS of MV3 extensions is sometimes flagged in the Chrome Web Store review process (https://github.com/InboxSDK/InboxSDK/discussions/957).

This change should not affect anyone except for users of the npm (non-remote) version of the InboxSDK who are using it in an MV2 extension and are using `InboxSDK.loadScript()` themselves. When they update to the latest InboxSDK, they will see an error message when they call `InboxSDK.loadScript()` recommending ~~they use a separate npm package that implements its old behavior~~ they use MV3. (Is anyone using the NPM build with MV2? Please speak up at https://github.com/InboxSDK/InboxSDK/discussions/1013)

Remaining work in this PR: Make a separate npm module for `InboxSDK.loadScript()` that affected users can switch to, and document it in the error message of the stub `InboxSDK.loadScript()` function.

Closes: #1011


src/inboxsdk-js/inboxsdk-NONREMOTE.ts
```diff
@@ -1,6 +1,8 @@
-import shimLoadScript from './loading/load-platform-implementation-DEV';
+import shimLoaderLoadScript from './loading/load-platform-implementation-NONREMOTE';
 import loader from './loading/platform-implementation-loader';
 
-loader._loadScript = shimLoadScript(500);
+loader._loadScript = shimLoaderLoadScript(500);
 
-export default require('./inboxsdk');
+import * as SDK from './inboxsdk';
```
wegry, line 6: I have a feeling the spec is being violated here. `import`, the keyword as opposed to the function `import()`, is meant to be statically resolved before the rest of the module loads (including line 4). Because of this, with one of the example extensions, it seems I'm seeing:

> loadScript is not available in the NPM version of the InboxSDK. This function is not usable in Chrome MV3 extensions. If you need this and you're not using an MV3 extension, use the NPM module <TODO> instead.

and

```
Uncaught ReferenceError: InboxSDK is not defined
    at content.js:5:1
```

To repro, try

```
yarn gulp clean
yarn gulp
```

with the thread-row example extension.

wegry: Removing the CJS (which is squirrelly) would be nice though if we can come up with an interface that works with ESM exclusively.

Macil: I think the preload method is called synchronously when the sdk module is loaded before our monkey-patch supplies which calls load script. I'm thinking of either moving the preload behind a setTimeout(..., 0) or making it be called right after we do the monkey-patch. I'll get this fixed soon. Sorry for pinging you on this before I put this through proper testing.

wegry: > I think the preload method is called synchronously when the sdk module is loaded which calls load script before our monkey-patch runs. I'm thinking of either moving the preload behind a setTimeout(..., 0) or making it be called right after we do the monkey-patch.


An alternative to `setTimeout` would be to lean on async a bit more. https://github.com/InboxSDK/InboxSDK/pull/971 shows how we might.

> Sorry for pinging you on this before I put this through proper testing.

No big.

src/inboxsdk.d.ts
```diff
@@ -15,18 +15,7 @@ import type { AppLogger } from './platform-implementation-js/lib/logger';
 export * from './platform-implementation-js/dom-driver/gmail/views/gmail-nav-item-view';
 
 export const LOADER_VERSION: string;
-export interface LoadScriptOptions {
```
wegry, line 18: We do depend on this interface in our downstream repo at Streak. Instead of removing this entirely, we might add an additional module `remote.d.ts` with this and TSDoc to the effect that:

> This interface only exists in the Manifest V2 (MV2) remote loaded version of the SDK. If and when MV2 support is dropped, this interface will be removed.

wegry:
I feel like we're close here. I think what's left here is

- merging #971 
- Testing against remote loading

wegry (approved):
I think we're ready to release this.

PR #971, opened 2023-08-07
Use deferred and promises to bootstrap the NONREMOTE SDK

wegry:
Uses the SDK's local Deferred and Promises to bootstrap the NONREMOTE SDK.

### DIFF
--- a/src/inboxsdk-js/inboxsdk-NONREMOTE.ts
+++ b/src/inboxsdk-js/inboxsdk-NONREMOTE.ts
@@ -1,8 +1,7 @@
-import shimLoaderLoadScript from './loading/load-platform-implementation-NONREMOTE';
-import loader from './loading/platform-implementation-loader';
-
-loader._loadScript = shimLoaderLoadScript(500);
-
+import { loadPi as shimLoaderLoadScript } from './loading/load-platform-implementation-NONREMOTE';
+import { PlatformImplementationLoader } from './loading/platform-implementation-loader';
 import * as SDK from './inboxsdk';
+
+PlatformImplementationLoader.loadScript = shimLoaderLoadScript(0);
 
 export default SDK;
--- a/src/inboxsdk-js/inboxsdk-NONREMOTE.ts
+++ b/src/inboxsdk-js/inboxsdk-NONREMOTE.ts
@@ -2,6 +2,6 @@ import { loadPi as shimLoaderLoadScript } from './loading/load-platform-implemen
 import { PlatformImplementationLoader } from './loading/platform-implementation-loader';
 import * as SDK from './inboxsdk';
 
-PlatformImplementationLoader.loadScript = shimLoaderLoadScript(0);
+PlatformImplementationLoader.loadScript = shimLoaderLoadScript(500);
 
 export default SDK;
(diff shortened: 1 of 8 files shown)


src/inboxsdk-js/loading/load-platform-implementation-NONREMOTE.ts
```diff
@@ -1,17 +1,9 @@
 import once from 'lodash/once';
 
-export default function (delay: number): () => Promise<void> {
```
wegry, line 3: The rub here is whether or not the delay way load bearing. With

```
  await new Promise(resolve => setTimeout(resolve, delay));
```

I couldn't get an example extension to start up.

wegry: Throwing this in and smoothing things should be something we do as part of this.

src/inboxsdk-js/inboxsdk-NONREMOTE.ts
```diff
@@ -1,8 +1,7 @@
-import shimLoaderLoadScript from './loading/load-platform-implementation-NONREMOTE';
-import loader from './loading/platform-implementation-loader';
-
-loader._loadScript = shimLoaderLoadScript(500);
-
+import { loadPi as shimLoaderLoadScript } from './loading/load-platform-implementation-NONREMOTE';
+import { PlatformImplementationLoader } from './loading/platform-implementation-loader';
 import * as SDK from './inboxsdk';
 
+PlatformImplementationLoader.loadScript = shimLoaderLoadScript(0);
```
wegry, line 5: This should have a timeout.

src/inboxsdk-js/loading/platform-implementation-loader.ts
```diff
@@ -1,28 +1,38 @@
...
-            if (!(global as any).__InboxSDKImpLoader) {
-              throw new Error('Implementation file did not load correctly');
-            }
-          });
-        }
-      })
-      .then(() => (global as any).__InboxSDKImpLoader.load('0.1', appId, opts));
-  },
+export class PlatformImplementationLoader {
+  static #loadScriptSet = defer<void>();
```
wegry, line 8: Using the deferred directly as the `loadScript` instead of as a flag would be a little bit more direct.

src/platform-implementation-js/main.ts
```diff
@@ -9,29 +16,27 @@ if (!global.__InboxSDKImpLoader) {
...
     } else {
-      resolve(
-        Kefir.merge([
-          Kefir.fromEvents(document, 'DOMContentLoaded'),
-          Kefir.fromEvents(window, 'load'),
-        ])
-          .take(1)
-          .map(() => null)
-          .toPromise()
+      const fromEvent = async (event: 'DOMContentLoaded' | 'load') =>
```
bkorobeinikov, line 26: i'm out of context what is this refactoring for, but flagging this behaviour change, when before first event that happened would trigger `resolve` is changed to wait untill all two events get triggered and then execute resolve. also i believe Kefir.fromEvents will unsubscribe when the take(1) is finished, but with new code it will never unsubscribe.

src/inboxsdk-js/inboxsdk-NONREMOTE.ts
wegry: Having toyed with this a bit, it _does_ seem that the timeout is load bearing. A timeout of zero doesn't seem to reliably load the SDK in the toy extension at examples/thread-rows.

src/platform-implementation-js/main.ts
wegry: [67259e3](https://github.com/InboxSDK/InboxSDK/pull/971/commits/67259e38c2c2900ea48997f7d6146a2a392d6e5b)

PR #974, opened 2023-08-08
Revert "chore: append sourceURL in lieu of sourceMappingURL"

wegry:
This reverts commit 958d1cde12a2e2552c41a20f0de7e7044df2dd57.

### DIFF
--- a/gulpfile.babel.ts
+++ b/gulpfile.babel.ts
@@ -255,7 +255,6 @@ async function webpackTask({
               // Adds hash to the end of the sourcemap URL so that
               // remote code source maps are not cached erroneously by Sentry.
               filename: '[file].map?hash=[contenthash]',
-              append: `\n//# sourceURL=https://www.inboxsdk.com/build/[file]?hash=[contenthash]`,
               publicPath: 'https://www.inboxsdk.com/build/',
             }),
           ]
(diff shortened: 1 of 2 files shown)


wegry:
When building the SDK with `yarn gulp --production --minify --remote` and [edc3d2b](https://github.com/InboxSDK/InboxSDK/pull/974/commits/edc3d2bfed40e45b829898818ffc5c82c23cb807), we end up with the following sourceMappingURLs:

| Output file | final line |
|--------|--------|
| inboxsdk.js | `//# sourceMappingURL=https://www.inboxsdk.com/build/inboxsdk.js.map?hash=0d31e9af422b79904b149543845c2e3c` |
| pageWorld.js | `//# sourceMappingURL=https://www.inboxsdk.com/build/pageWorld.js.map?hash=f9d56f923630039220e2d943546c08a3` |
| platform-implementation.js | `//# sourceMappingURL=https://www.inboxsdk.com/build/platform-implementation.js.map?hash=3050f525b7c89e7c481baef1a7f8fcb2` | 

# Remaining issues

I believe we'll still end up with both a `//# sourceMappingURL` and `//# sourceMapURL` comment (added at runtime in our downstream app) even after this change. 

- #960 doesn't seem to have taken effect @Macil 
- There may be something around our `eval` usage for loading scripts preventing `//#sourceMappingURL` from being picked up by sentry.

PR #976, opened 2023-08-08
docs(#939): clarify around ThreadRow/MessageView interaction with TSDoc

### LINKED ISSUE #939: MessageView.getSender() throws an error for hidden messages in thread
**Describe the bug**

MessageView.getSender() throws an error for hidden messages in thread:
* `Failed to find element with selector: td.gF span[email]`

**To Reproduce**

1. Great email thread with 6 messages
2. Register thread button
3. Try to get list of senders on thread button click

```typescript
InboxSDK.load(2, "Hello World!").then((sdk) => {
  sdk.Toolbars.registerThreadButton({
    title: 'Start',
    positions: ['THREAD'],
    iconUrl: StartImage,
    hasDropdown: true,
    onClick(event) {
      const selectedThreadViews = event.selectedThreadViews;
      const messages: InboxSDK.MessageView[] = selectedThreadViews.reduce(
          (acc, view) => acc.concat(view.getMessageViewsAll()),
          [],
      );
      const senders: string[] = messages.map((message) => message.getSender().emailAddress);

     // ISSUE when there are hidden messages in tread: MessageView.getSender() fails: `Failed to find element with selector: td.gF span[email]`

      onClick({
        root: event.dropdown.el,
        emails: [...new Set(senders)],
        onClose: () => event.dropdown.close(),
      });
    },
  });
});


```

**Expected behavior**

* `MessageView.getSender()` should return `null` OR `undefined` if the sender email cannot be found in DOM instead of throwing exception

**Screenshots**

image
image

wegry:
Colocate docs relevant to #939 with the ThreadRow MessageView interface.

Closes #939


Macil (changes requested):
Multiple methods on MessageView are expected to error if the MessageView isn't loaded yet. If you use a function that specifically retrieves unloaded MessageViews (like ThreadView.getMessageViews*All*), you're expected to check the MessageView.isLoaded method yourself to determine if you can interact with the MessageView. Instead of changing getSender like this, we should update the docs to be clearer about the existing behavior.

src/platform-implementation-js/namespaces/conversations.ts
```diff
@@ -6,9 +6,18 @@ import AttachmentCardView from '../views/conversations/attachment-card-view';
...
 const memberMap = new WeakMap();
+
 export const MessageViewViewStates = Object.freeze({
+  /**
+   * One case where a {@link MessageView} is 'HIDDEN' is when there are too many messages in a thread,
+   * and the message doesn't show at all.
+   */
   HIDDEN: 'HIDDEN',
+  /**
+   * {@link MessageView} are collapsed when they are partially shown. The {@link MessageView}'s subject line and timestamp is visible, but the body of the message may be truncated.
```
wegry, line 19:
```suggestion
   * {@link MessageView}s are collapsed when they are partially shown. The {@link MessageView}'s subject line and timestamp is visible, but the body of the message may be truncated.
```

Issue #942, opened 2023-07-06
Failed to find element with selector: td div.yW


wegry:
Splitting this out from https://github.com/InboxSDK/InboxSDK/issues/941#issuecomment-1622903112.

```
Failed to find element with selector: td div.yW
../../../node_modules/@inboxsdk/core/inboxsdk.js in new GmailActionButtonView at line 8917:31
../../../node_modules/@inboxsdk/core/inboxsdk.js in getTitleHTMLString at line 20434:95
../../../node_modules/@inboxsdk/core/inboxsdk.js in new GmailActionButtonView at line 8917:31
../../../node_modules/@inboxsdk/core/inboxsdk.js in GmailAttachmentCardView._createNewElement at line 13761:38
../../../node_modules/@inboxsdk/core/inboxsdk.js in s0._add at line 41020:1
../../../node_modules/@inboxsdk/core/inboxsdk.js in s0._handleMain at line 41020:1
../../../node_modules/@inboxsdk/core/inboxsdk.js in _$handleMain at line 41020:1
../../../node_modules/@inboxsdk/core/inboxsdk.js in L at line 41020:1
../../../node_modules/@inboxsdk/core/inboxsdk.js in F.dispatch at line 41020:1
../../../node_modules/@inboxsdk/core/inboxsdk.js in ve._emitValue at line 41020:1
../../../node_modules/@inboxsdk/core/inboxsdk.js in ve._onTick at line 41020:1
../../../node_modules/@inboxsdk/core/inboxsdk.js in _$onTick at line 41020:1
../../../node_modules/@sentry/browser/node_modules/@sentry/utils/esm/envelope.js in serializeEnvelope at line 83:43
../../../node_modules/@inboxsdk/core/inboxsdk.js in _getChipElement at line 11927:111
```

wegry:
At Streak, we have [1.5K instances of this](https://streak.sentry.io/issues/?environment=production&groupStatsPeriod=auto&page=0&project=4504322969763840&query=%22Failed+to+find+element+with+selector%3A+td+div.yW%22&referrer=issue-list&sort=user&statsPeriod=14d) `Error` occurring in the last two weeks.

wegry:
[...]

PR #993, opened 2023-08-15
chore: check places isUsingSyncAPI variants are used

wegry:
Digging into [isUsingSyncAPI](https://github.com/InboxSDK/InboxSDK/blob/c8b9f02934d8cd87121bb121277bed046365144a/src/platform-implementation-js/dom-driver/gmail/gmail-page-communicator.ts#L146) for #942, it appears that isUsingSyncAPI=true is a million times more frequently logged than false in the last week up until 2023-08-15. Ignoring (or removing) isUsingSyncAPI=false seems like it might be a viable path forward


wegry:
[Streak internal link](https://docs.google.com/spreadsheets/d/1UVcFqgXXfWPsbfIO-s-LvAYSrCtjhOoUFwn1ZY9R5zM/edit?usp=sharing) showing logged `isUsingSyncAPI` over the last week.

| isUsingSyncAPI | c |
| --- | ---: |
| true | millions |
| null | 5 |
| false | 3 |

Issue #787, opened 2022-03-08
NavMenu and supportItem issue.


shuhankuang:

wegry:
We have at least ~8K instances of this error occurring in the last 14D at Streak ([internal link](https://streak.sentry.io/discover/homepage/?field=title&field=event.type&field=project&field=user.display&field=timestamp&id=20561&interval=1d&name=&project=4504322969763840&query=shouldAddNavItemsInline+failed+to+find+leftNavElement&sort=-timestamp&statsPeriod=14d&topEvents=5&yAxis=count%28%29)).

[The way this error was added](https://github.com/InboxSDK/InboxSDK/pull/713) hints it's probably occurring in an older layout configuration of Gmail.

wegry:
Hey @shuhankuang, thanks for the report. I'm going to create a new issue for the second part of your issue. We should focus on 

```
Uncaught (in promise) Error: shouldAddNavItemsInline failed to find leftNavElement
    at Object.shouldAddNavItemsInline (inboxsdk.js:26:15809)
```

in this issue.

PR #996, opened 2023-08-16
types: connect legacy NavMenu namespace down to shouldAddNavItemsInline

wegry:
Poking at #787 with tsc.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
@@ -153,15 +153,12 @@ const GmailElementGetter = {
     return document.body.querySelector('.dw .nH > .nH > .no');
   },
 
-  // This function checks whether we should use the old InboxSDK style of adding nav items
-  // inline among Gmail's nav items (as opposed to the newer style where we put our nav items
-  // in their own sections at the bottom of the leftnav).
+  /**
+   * This method checks whether we should use the old InboxSDK style of adding nav items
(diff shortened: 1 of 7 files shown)


src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
```diff
@@ -175,14 +172,6 @@ const GmailElementGetter = {
 
     // leftNavElement classnames depending on gmail chat & meet settings:
 
-    // Old Hangouts-only Gmail
```
wegry, line 178: Screenshot 2023-08-16 at 13 25 06

It's my understanding that hangouts support in Gmail is no more. Maybe Chat holds onto these older class names?

src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
```diff
@@ -175,14 +172,6 @@ const GmailElementGetter = {
...
-    //  div.aeN.WR.nH.oy8Mbf.nn
-    //  (looks identical to current Gmail's chat-off-meet-off setup, but there are important
-    //  CSS differences inside the element!)
-    // ===================
-    // No classname difference between open/collapsed
-    //  div.ajl.aib.aZ6 (classic hangouts on left or right, google meet on/off)
```
wegry, line 184: https://support.google.com/chat/answer/9341104?hl=en&co=GENIE.Platform%3DDesktop suggests otherwise, but chat position as a setting seems to not exist as it did in hangouts. 

Screenshot 2023-08-16 at 13 28 58

PR #1015, opened 2023-09-04
Added data-tooltip to Toolbar Button

mohsinaliryk:
The tooltip was not displaying when hovering the thread row view button. Upon inspecting the Gmail DOM, it was observed that other native Gmail buttons utilize the data-tooltip attribute for tooltip text.


aleemstreak:
Thanks for submitting this fix! We'll take a quick look tomorrow then merge and cut a release.

PR #1017, opened 2023-09-05
added missing title property

mohsinaliryk:
`title`  property was missing when passing Button descriptor to  `gmailThreadRow.addButton()`


wegry (commented):
As this is, I'd like to thread through typing from the top interface. I'll take a closer look on Monday at the earliest.

wegry (approved):
After playing with the types a bit, I think they're sufficiently convoluted we can just merge this. Thanks for this.

PR #1027, opened 2023-09-15
Add User.getPersonDetails()

Macil:
Adds a new `sdk.User.getPersonDetails(emailAddress: string): Promise<PersonDetails | undefined>` function where `PersonDetails` is

```ts
export interface PersonDetails {
  fullName?: string;
  fullNameSortOrder?: string;
  firstName?: string;
  lastName?: string;
  imageUrl?: string;
}
```

This is used so we can get the avatar image from Gmail from a user's email address. In Streak we use this when we the user views an email from a teammate's inbox, and we want to render the email including the avatar as it would be rendered in Gmail natively.


jfmaggie (approved):
lgtm

src/platform-implementation-js/namespaces/user.ts
```diff
@@ -1,59 +1,64 @@
 import find from 'lodash/find';
-import * as ud from 'ud';
-import get from '../../common/get-or-fail';
 import type { Driver } from '../driver-interfaces/driver';
 import type { PiOpts } from '../platform-implementation';
-const memberMap = ud.defonce(module, () => new WeakMap()); // documented in src/docs/
```
wegry, line 6: #910 touched on this a bit, but has something changed around using `ud` v. class members?

Macil: I was uncomfortable with doing a blanket rip of ud, but I'm fine with it being taken out anywhere it gets in the way of something (like moving away from these weakmap standins for true private properties).

wegry: I mimicked this [here](https://github.com/InboxSDK/InboxSDK/pull/1029/files#diff-36c20528977bbd53e3b13505ebf5a0c07cd1833807dbb66f70323df5b95e3bd0R15).

PR #1032, opened 2023-09-22
fix(#1030): accept `T | Observable<T, unknown>` in TS for addSidebarContentPanel

### LINKED ISSUE #1030: Bug: `addSidebarContentPanel()` parameter type changed in `2.1.2`
When updating from `2.2.1` to `2.1.2` the argument type of `addSidebarContentPanel(descriptor)` changed from `ContentPanelDescriptor` to `Observable<ContentPanelDescriptor, unknown>`.

I expected this type not to change, especially within a patch update.

Is this intended? If yes, how should we do now? (perhaps the docs should be updated as well).

Thank you!

wegry:
Closes #1030

### DIFF
--- a/src/inboxsdk.d.ts
+++ b/src/inboxsdk.d.ts
@@ -406,7 +406,9 @@ export interface SimpleElementView extends EventEmitter {
 export interface ThreadView extends TypedEventEmitter<ThreadViewEvents> {
   addLabel(): SimpleElementView;
   addSidebarContentPanel(
-    contentPanelDescriptor: ContentPanelDescriptor,
+    contentPanelDescriptor:
+      | ContentPanelDescriptor
+      | Kefir.Observable<ContentPanelDescriptor, unknown>,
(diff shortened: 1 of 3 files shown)


wegry:
@sybers if you don't mind trying this out locally against tsc, I'd appreciate it.

To build the SDK locally and try this out:

```sh
# change into SDK directory after cloning
yarn gulp

# in your project using the SDK
yarn link [path to SDK repo]/packages/core    # might be npm or pnpm depending on your project
```

sybers:
Hey @wegry, I checked out this PR locally and linked it to our project and it looks like it's working fine now :)

Another thing I noticed is that the current return type of `addSidebarContentPanel() ` is `Promise<ContentPanelView | null | undefined>`, however there are no branches which return `undefined`. The return type could be simplified to `Promise<ContentPanelView | null>`.

Macil (approved):

PR #1036, opened 2023-10-02
fix: sidebar mole position when chat enabled

wegry:
- relax type def around ContentPanelDescriptor
- enforce private on ViewDriver fields/methods
- fix: keep track of order property for moles added

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/widgets/gmail-mole-view-driver.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/widgets/gmail-mole-view-driver.ts
@@ -12,47 +12,47 @@ import isComposeTitleBarLightColor from '../is-compose-titlebar-light-color';
 import * as styles from './mole-view.module.css';
 
 class GmailMoleViewDriver {
-  _driver: GmailDriver;
-  _eventStream = kefirBus<
+  #driver: GmailDriver;
+  #eventStream = kefirBus<
     {
       eventName: 'minimize' | 'restore';
     },
     unknown
   >();
-  _stopper = kefirStopper();
-  _element: HTMLElement;
+  #stopper = kefirStopper();
+  #element: HTMLElement;
 
   constructor(driver: GmailDriver, options: MoleOptions) {
-    this._driver = driver;
-    this._element = Object.assign(document.createElement('div'), {
+    this.#driver = driver;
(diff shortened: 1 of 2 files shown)


src/platform-implementation-js/dom-driver/gmail/widgets/gmail-mole-view-driver.ts
```diff
@@ -97,14 +97,32 @@ class GmailMoleViewDriver {
...
+         * If we don't, we end up with moles added on top of the sidebar because of a chat element in the same parent element as moleParent.
+         *
+         * This fix ratches the order of the element padding in order to present moles before the sidebar. Whether or not this fix works with Gmail's CSS in a RTL language is unknown.
+         */
+        const oldOrder = lastChild.style.order;
+        const oldOrderNum = parseInt(oldOrder, 10);
+
+        if (oldOrder && !isNaN(oldOrderNum)) {
+          this.#element.style.order = oldOrder;
+          lastChild.style.order = `${oldOrderNum - 1}`;
```
dmy7r0, line 121: Q: Does this mean we'll end up with duplicate order number elements?

dmy7r0 (commented):
The approach looks smart. Had a question regarding the order attribute

src/platform-implementation-js/dom-driver/gmail/widgets/gmail-mole-view-driver.ts
wegry: A: no, because the last child (the spacer on the right of the composes):

- starts at 0
- Gmail seems to only use positive integers for compose and chat moles `order` properties

The last bullet point means that this fix causes any mole that's added to always be to the right of any subsequently added compose. That probably dooms this fix.

wegry:
When

1. Programmatically adding an SDK mole
2. Adding a compose
3. Programmatically adding another SDK mole

[3b12717](https://github.com/InboxSDK/InboxSDK/pull/1036/commits/3b12717fbd20a9f78fcdc7df8c74cae0cfb00d5d) will cause adding them to be ordered from left to right as `[2] [1] [3]` instead of `[3] [2] [1]`.

There may be a way to trick Gmail's JS into recognizing the order of the SDK mole prior and modifying its order.

PR #1051, opened 2023-10-31
fix: stop compose from flashing in if SDK mole's moleParent doesn't exist yet

wegry:
When we add SDK moles to Gmail, we add them in the same element that Gmail adds ComposeViews (and if Chat is enabled, Chat moles)--the moleParent. The moleParent doesn't seem to exist by default when Gmail starts up unless:

- there's a ComposeView lingering from a previous Gmail load
- The user is on a route like MessageView where there are inline ComposeViews

The way we ensure the moleParent is present (only an issue once per page load) is by quickly opening a ComposeView and then closing it. This can be slow enough that there is a FOUC around the ComposeView showing up alongside an SDK mole.

This PR adds a class to `document.body` while we're forcing moleParent into existence to hide the FOUC. It also discards instead of just closing the temporary compose draft to handle the edge case for _users with email signatures where flashing in a compose inadvertently saves a draft_.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
@@ -149,10 +149,6 @@ const GmailElementGetter = {
     return getMainContentElementChangedStream(this);
   }),
 
-  getMoleParent(): HTMLElement | null {
-    return document.body.querySelector('.dw .nH > .nH > .no');
-  },
-
   /**
    * This method checks whether we should use the old InboxSDK style of adding nav items
    * inline among Gmail's nav items (as opposed to the newer style where we put our nav items
--- a/src/platform-implementation-js/dom-driver/gmail/widgets/gmail-mole-view-driver.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/widgets/gmail-mole-view-driver.ts
@@ -298,7 +298,7 @@ class GmailMoleViewDriver {
           document.body.classList.add(styles.hideComposes);
           const gmailComposeView =
             await this.#driver.openNewComposeViewDriver();
-          gmailComposeView.close();
+          gmailComposeView.discard();
         } finally {
           document.body.classList.remove(styles.hideComposes);
         }
--- a/src/platform-implementation-js/dom-driver/gmail/widgets/gmail-mole-view-driver.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/widgets/gmail-mole-view-driver.ts
@@ -1,7 +1,6 @@
 import * as Kefir from 'kefir';
 import kefirBus from 'kefir-bus';
 import kefirStopper from 'kefir-stopper';
-import streamWaitFor from '../../../lib/stream-wait-for';
 import querySelector from '../../../lib/dom/querySelectorOrFail';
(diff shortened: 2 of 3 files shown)


aleemstreak:
why not discard the compose instead of just closing it? that will prevent the saving of the draft

wegry:
> why not discard the compose instead of just closing it? that will prevent the saving of the draft

@aleemstreak Changed in e855b379f6aae3e2671d3eb3f7bc9c16fac9857d. @ruthhalteman might weigh in testing though. I haven't had much luck reproing the drafts being saved.

aleemstreak:
dumb question, but what if we just add the moleParent ourselves instead of opening a compose? or does gmail need to do it for its own state?

wegry:
> dumb question, but what if we just add the moleParent ourselves instead of opening a compose? or does gmail need to do it for its own state? - @aleemstreak

Gmail seems to have its own reference to the moleParent in its JS state. 

As an experiment: with the following diff on top of this PR

```diff
diff --git a/src/platform-implementation-js/dom-driver/gmail/widgets/gmail-mole-view-driver.ts b/src/platform-implementation-js/dom-driver/gmail/widgets/gmail-mole-view-driver.ts
index 9b2c1468..d75cf4a3 100644
--- a/src/platform-implementation-js/dom-driver/gmail/widgets/gmail-mole-view-driver.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/widgets/gmail-mole-view-driver.ts
@@ -293,15 +293,25 @@ class GmailMoleViewDriver {
       .takeUntilBy(this.#stopper)
       .onValue(async () => {
         this.#driver.getLogger().eventSdkActive('mole parent force load');
-
-        try {
-          document.body.classList.add(styles.hideComposes);
-          const gmailComposeView =
-            await this.#driver.openNewComposeViewDriver();
-          gmailComposeView.discard();
-        } finally {
-          document.body.classList.remove(styles.hideComposes);
-        }
+        console.log('APPENDING');
+
+        const sdkMoleParent = document.createElement('div');
+        sdkMoleParent.classList.add('dw', 'np');
+        sdkMoleParent.innerHTML = `
+          <div class="dw np">
+            <div>
+              <div class="nH">
+                <div class="nH">
+                  <div class="no" style="float: right">
+                    <div class="nH nn" style="order: 2147483647; width: 56px"></div>
+                    <div class="nH nn" style="order: 0; width: 66px"></div>
+                  </div>
+                  <div class="dJ"></div>
+                </div>
+              </div>
+            </div>
+          </div>`;
+        document.body.appendChild(sdkMoleParent);
       });
   }
```

1. Open Gmail to the inbox view. `$('.dw')` in the browser console should return `null`.
2. Open an SDK mole to trigger the block above. `$('.dw')` should now return the node added by the SDK.
3. Open a Gmail Compose mole. `document.querySelectorAll('.dw')` unfortunately returns two elements, the SDK added moleParent and the Gmail one.

Example state:
Screenshot 2023-11-01 at 08 44 27

wegry:
@ruthhalteman found an issue during testing where after this change, Streak's extension doesn't appear be able show any SDK moles if Google Chat is enabled.

Digging in to see why our temporary `composeView`'s `composeView.discard()` call throws in that case.

ruthhalteman (approved):

PR #1067, opened 2023-11-14
log html if thread view toolbar was not found

bkorobeinikov:
Some Streak user's reporting issue with buttons not being added into the thread view toolbar (via `registerThreadButton`). Looks like Gmail is rolling out a gradual update. 

log censored html of the thread view container if toolbar wasn't found or fallback to log censored page html if thread view container wasn't found.


aleemstreak:
lgtm, ship it.

we can ask users to refresh who are having the issue

Issue #1072, opened 2023-11-16
Failed to find element with selector: `table.Bs > tr`


kuhnster01:
Hi there, we are seeing this error in the console for users in Australia (but Gmail is potentially rolling this out further.  It may or may not be related to emails opening (as we have a separate issue) but we want to know if anyone else is seeing this InboxSDK specific
issue:

Sent by App: false 
Session Id: 1700115147601-0.4718526429030505 
Extension Id: chrome-extension://aohlfneeliakfcefeffppfplagbccbni/ 
InboxSDK Loader Version: 0.7.24-1484787998857-c248fbb55be579d3 
InboxSDK Implementation Version: 2.1.13-1699993741673-7af9b85c39a0158d
eval @ platform-implementation.js:23
Show 1 more frame
Show less
error-handler.js:192 2023-11-16T06:12:43.872Z: {"level":"error","message":"Logging chrome extension error to the server.","meta":{"message":"Logging chrome extension error to the server.","technical":"Failed to find element with selector: table.Bs > tr","originalError":{"message":"Failed to find element with selector: table.Bs > tr","stack":"Error: Failed to find element with selector: table.Bs > tr\n    at r (https://www.inboxsdk.com/build/platform-implementation.js:13:61559)\n    at He._startMonitoringPreviewPaneForThread (https://www.inboxsdk.com/build/platform-implementation.js:79:10452)\n    at He._setupContentAndSidebarView (https://www.inboxsdk.com/build/platform-implementation.js:79:10063)\n    at eval (https://www.inboxsdk.com/build/platform-implementation.js:79:9370)\n    at l.call (https://www.inboxsdk.com/build/platform-implementation.js:2:3304)\n    at MutationObserver.a (https://www.inboxsdk.com/build/platform-implementation.js:2:3550)\n    at v.invoke (chrome-extension://aohlfneeliakfcefeffppfplagbccbni/components/polyfills.js:1:6566)\n    at M.runGuarded (chrome-extension://aohlfneeliakfcefeffppfplagbccbni/components/polyfills.js:1:2085)\n    at MutationObserver.<anonymous> (chrome-extension://aohlfneeliakfcefeffppfplagbccbni/components/polyfills.js:1:1864)"},"logOnly":true,"time":"2023-11-16T06:12:43.872Z","ceVersion":"2.2.459.11749"},"name":"LogMessage"}
polyfills.js:1 Uncaught Error: Failed to find element with selector: table.Bs > tr
[...]

Issue #1071, opened 2023-11-16
`Conversations.registerThreadViewHandler` not triggered with split pane enabled


Alex-Chopard:
Hi, I've found another bug that seems related to a Gmail update, but, not all account are impacted by it.

**Describe the bug**
`inboxSDK.Conversations.registerThreadViewHandler` callback isn't called if the split pane mode is on and the list is loaded from a search.

This bug is similar to: https://github.com/InboxSDK/InboxSDK/issues/1042

**To Reproduce**

1. Have an affected account
2. Load the extension [thread-test](https://github.com/InboxSDK/InboxSDK/tree/main/examples/thread-test)
3. Enabled the split pane mode
4. On `#inbox` everything works well
5. After a search (e.g. #search/foo) the `registerThreadViewHandler` callback isn't called anymore

**Expected behavior**
Registered callback should be called.

**Screenshots**
Without a search (#label/test), the [log](https://github.com/InboxSDK/InboxSDK/blob/main/examples/thread-test/content.js#L47C10-L47C10) `threadView` appear in the console
![Screenshot 2023-11-16 at 11 53 19](https://github.com/InboxSDK/InboxSDK/assets/29271954/590ecbfb-ec40-45fe-81f8-8f32dfedf0ad)

With a search (#search/label:test google)), the [log](https://github.com/InboxSDK/InboxSDK/blob/main/examples/thread-test/content.js#L47C10-L47C10) `threadView` does not appear in the console
![Screenshot 2023-11-16 at 11 56 31](https://github.com/InboxSDK/InboxSDK/assets/29271954/7d009330-5507-4f3e-8101-d745487045b9)

**Add any other context about the problem here if anything**
I've look a bit in the code to try to find where and why that bug appear, and it's seems due to a selector that is called too early. When I have the bug, this [selector](https://github.com/InboxSDK/InboxSDK/blob/main/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts#L346) return null.

wegry:
> I've look a bit in the code to try to find where and why that bug appear, and it's seems due to a selector that is called too early. When I have the bug, this [selector](https://github.com/InboxSDK/InboxSDK/blob/main/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts#L346) return null.

@Alex-Chopard It looks like you own an account that can repro that selector failure. Does the selector appear later?

[...]

Issue #1070, opened 2023-11-16
Bug: Unable to open right panel for some customer on new Gmail  version


fstn:
When opening the right Panel:

```
    index.8b6ab181.js:55 Error logged: Error: Failed to find element with selector: table.Bs > tr
    at c (index.8b6ab181.js:145:13093)
    at l8._startMonitoringPreviewPaneForThread (index.8b6ab181.js:226:4867)
    at l8._setupContentAndSidebarView (index.8b6ab181.js:226:4481)
    at index.8b6ab181.js:226:3793
    at z.call (index.8b6ab181.js:112:3465)
    at MutationObserver.x (index.8b6ab181.js:112:3706)
 ```

When opening a sent email:

<img width="1334" alt="image" src="https://github.com/InboxSDK/InboxSDK/assets/2936857/f1426bf7-36dc-4d08-8be5-3cd94b0d017c">

"@inboxsdk/core": "^2.0.1", and tried last 2.1.13 
Macbook
Chrome Version 119.0.6045.159 (Official Build) (arm64)

Regard

wegry:
Hey @fstn, thanks the report. 

#1073, released as part of `@inboxsdk/core@2.1.15`, might help with this issue.

wegry:
Closing as dupe of #1072.

Issue #1068, opened 2023-11-16
Bug: 


FrancescoBB:
**Frequency**: only for a limited number of users among the ones using the extension which makes use of Inbox SDK, but for those users the issue is always happening (no random behavior)

**Version**: tested with the last one available (see screenshots below)

**Issue Description**: when loading components related to a single email by using the methods

1- sdk.Toolbars.registerToolbarButtonForList
2- sdk.Conversations.registerMessageViewHandler

the button and the sidebar are not loaded at all and the two errors in the screenshots below appears (sometimes the first one, sometimes the second one, depending on which mail is selected, with no apparent pattern dictating which one happens when)

![image](https://github.com/InboxSDK/InboxSDK/assets/62301150/beaf312b-2046-4ce5-9a8e-3c9479a36cb6)
![image](https://github.com/InboxSDK/InboxSDK/assets/62301150/10d06466-352d-4bc8-8e03-8be001501c1d)

**Expected behavior**:

Buttons and sidebar appearing correctly as it happens with most of the users

Other Notes:

1- Issue started on 2023-11-14
2- All the users of the workgroup belongs to the same gmail domain 
3- The issue happens both in normal mode and in delegated mode (while the not affected users are ok both in normal and delegated mode)
4- No relavant differences in the gmail settings among the users

aleemstreak:
Do you have access to an account that is exhibiting this behaviour?
ᐧ

FrancescoBB:
@aleemstreak yes, it's an account used by one of our customers and I've been able to reproduce the issue yesterday by using the user's account.

aleemstreak:
is that something you could share with us?

FrancescoBB:
@aleemstreak no, unfortunately I cannot share the account as it is the customer's private one.
But I can access it and take some logs if it can be of help for you.
In case, which ones would you need?

wegry:
@FrancescoBB #1073 may help with the issue.

FrancescoBB:
Thanks @wegry we'll give it a try asap!

FrancescoBB:
@wegry the fix you referenced solved the problem for our user!
Many thanks for the fast intervention!

PR #1073, opened 2023-11-16
selectors update for gmail thread view container changes

bkorobeinikov:
gmail started rolling out update to its thread view markup, new markup no longer contains `table.Bs` selector, PR switched to use `.a98.iY` selector (because it is present in old and new markup, it required extra condition changes etc)

image

the fix is tentative and may need to be tweaked further.


wegry:
Related: #1068, #1070, #1071, and #1072.

wegry (approved):

PR #1079, opened 2023-11-17
support extracting whole document html and css

bkorobeinikov:
because of gmail gradual updates there is a need in logging full document html with css so then it is possible to inspect the DOM tree and identify changes.

`extractDocumentHtmlAndCss` will output obfuscated document.body content and all css that is used on the page

example of the output
image

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
@@ -24,8 +24,8 @@ import { makePageParser } from './page-parser';
 import toItemWithLifetimeStream from '../../../../lib/toItemWithLifetimeStream';
 import waitFor from '../../../../lib/wait-for';
 import { SelectorError } from '../../../../lib/dom/querySelectorOrFail';
-import censorHTMLstring from '../../../../../common/censorHTMLstring';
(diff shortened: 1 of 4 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
```diff
@@ -394,7 +394,7 @@ class GmailRouteView {
       });
       if (isStreakAppId(this._driver.getAppId())) {
         this._driver.getLogger().error(selectorError, {
-          html: censorHTMLstring(previewPaneContainer.innerHTML),
+          html: extractDocumentHtmlAndCss(),
```
aleemstreak, line 397: this is going to log a lot more data to our servers - do you know the rough frequency of these errors? Just want to make sure we don't DDOS ourselves.

aleemstreak: if its high frequency, we can sample

bkorobeinikov: @Macil pointed out the the rate limit implemented for errors https://github.com/InboxSDK/InboxSDK/blob/d2ea8b878c95570204e2d2b9db2c95d89e379c1e/src/common/log-error.ts#L197

Issue #1078, opened 2023-11-17
`Compose.registerComposeViewHandler` Registered Compose View Handler does not fire on Reply/Reply All/Forward


emilebourquin:
This is probably related to the Gmail changes that occurred yesterday, as this behavior started yesterday as well, in a small number of Gmail accounts. We are running the latest InboxSDK 2.1.15

Our extension "pops-out" Compose, Reply, Reply-All, and Forwards. On a small number of Gmail accounts, the Reply/Reply All/Forward threads no longer auto-pop-out. Compose does still auto-pop-out.

I can see that the Registered Compose View Handler does not fire on Reply/Reply All/Forward, so I'm guessing the selector is incorrect.

Below are two images of Chrome Dev Tools searching for .a98.iY . The selector is found in both working and non-working, but the surrounding DOM elements are quite different between the two:

**Works properly:**
![EmileWorks a98 iY](https://github.com/InboxSDK/InboxSDK/assets/89332/96c629af-8335-44b6-b3a9-e079597a3491)

**Does not work:**
![TimDoesNotWork a98 iY](https://github.com/InboxSDK/InboxSDK/assets/89332/f2273a1d-1e63-416a-a656-115395811c66)

Hacking on this a bit, it seems the selector in the Not Working scenario should be '.nH.UG.id:has(.a98.iY)' instead of '.nH.g.id:has(.a98.iY)' , but I still can't get the Compose View Handler to fire.

wegry:
Hey @emilebourquin, thanks for the report.

Mind creating a minimal repro extension? Our https://github.com/InboxSDK/InboxSDK/tree/main/examples folder might help you get started.

emilebourquin:
Thanks @wegry, the 2.1.16 update seems to have (mostly?) fixed this issue. Occasionally, in the problem Gmail account, we're getting the "Thread container for preview pane wasn't found" error. If it happens again, I'll post some detail.

I'll also go down the road of getting a minimal extension set up to reproduce any possible issues.

emilebourquin:
@wegry were still seeing issues intermittently where the "Thread container for preview pane wasn't found" error is thrown, and when the Reply does not auto-pop-out the ".ao8:has(.a98.iY)" selector is found three times in the DOM, so perhaps it's not specific enough?

emilebourquin:
[...]

PR #1081, opened 2023-11-20
fix order of thread view parsing in split mode

bkorobeinikov:
`.querySelector` will return the second match if it is within children of the first match, so in order to keep inboxsdk expected container, the selectors should be split


wegry (approved):
LGTM. I noticed the previous version with the bundled query selector in gmail-route-view was sometimes giving back `.ao8` inadvertently when `table.Bs > tr` was present.

As an aside, I tried this against #1078, and it doesn't seem to fix the issue.

PR #1083, opened 2023-11-21
types: ThreadView and RowListView in Driver

wegry:
Improves typing in Driver around Toolbars, RouteView, and ThreadView.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/gmail-driver.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/gmail-driver.ts
@@ -101,6 +101,7 @@ import type {
   SearchSuggestionsProvider,
   SearchQueryRewriter,
 } from '../../namespaces/search';
+import isNotNil from '../../lib/isNotNil';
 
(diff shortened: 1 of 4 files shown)


src/platform-implementation-js/dom-driver/gmail/gmail-driver.ts
```diff
@@ -827,37 +825,42 @@ class GmailDriver {
...
+      Kefir.merge<GmailToolbarView | null | undefined, unknown>([
+        this.#rowListViewDriverStream.map((gmailRowListView) =>
+          gmailRowListView.getToolbarView(),
         ),
         this.getThreadViewDriverStream().map((gmailThreadView) =>
           gmailThreadView.getToolbarView(),
         ),
       ])
-        .filter(Boolean)
+        .filter(isNotNil)
```
wegry, line 861: This should be the only runtime change in this PR. `isNotNil` has a type assertion that feeds into the rest of the method chain.

PR #1084, opened 2023-11-21
support gmail page loading state

bkorobeinikov:
it appears that the Gmail search page now renders with a delay, first showing a loading spinner (and potentially more pages will be doing that soon). This broke the navigation handler that expects elements to be present instantly. Gmail adds `[role=main]` attribute once loading is finished.

image

the issue was reproduced on the account provided by @Alex-Chopard and @RBische thank you!

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
@@ -311,11 +311,9 @@ class GmailRouteView {
       // role=main attribute is not set while page in a loading state
       await waitFor(() => document.querySelector('[role=main]'), 15_000);
     } catch {
-      this._driver
-        .getLogger()
-        .error(new Error('[role=main] element not found'), {
-          html: extractDocumentHtmlAndCss(),
-        });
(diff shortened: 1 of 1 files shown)


wegry (approved):
In my local testing, this looks like it patches up #1071 on the provided account.

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
```diff
@@ -293,15 +295,30 @@ class GmailRouteView {
...
   }
 
+  async #waitForMainElementSafe() {
+    try {
+      // role=main attribute is not set while page in a loading state
+      await waitFor(() => document.querySelector('[role=main]'), 15_000);
+    } catch {
+      this._driver
+        .getLogger()
+        .error(new Error('[role=main] element not found'), {
```
wegry, line 316: We might use `SelectorError` here instead of `Error`.

bkorobeinikov: changed

vishnug-bridge:
This issue still exist... i am using v2.1.26

Issue #798, opened 2022-10-20
Dealing with Gmail breaking changes without remote code loading (MV3)


Macil:
With Chrome MV3, we can't remote-load code, which removes our previous way of quickly reacting to Gmail changes that break InboxSDK compatibility. This is an issue tracking some of our plans to work on making InboxSDK resilient in the face of future Gmail changes that might surprise us and break certain functionality.

- Fallbacks for navmenu, custom views, thread sidebar, etc, so if we fail to find the elements to inject these features into, we have a less-integrated but dependable alternative implementation (for navmenu/sidebar, a position:fixed toggleable pop-in available at the side of the screen, and for custom views a position:fixed modal).
- Remote-loaded config that includes:
  - CSS selectors to use to find elements (and [page-parser-tree](https://github.com/StreakYC/page-parser-tree) configurations)
  - object paths in request interception for email send events
  - whether to force fallbacks (possibly conditional on specified CSS selectors matching)
  - selectors to log censorHTMLTree() results on to help debug unseen Gmail HTML structure
  - console logging messages to developers when they're using a critically outdated InboxSDK version

The remote-loaded config will be something hosted by us. We can add options to allow this config to be disabled or fetched from an alternative URL in case some extension developers want to be able to manage it on their own.

ben1lam8:
Really interesting. It's so cool you decided to maintain a server that would help us all. 🙏🏻

designcode:
@Macil Any updates on this one? We would like to move to V3 ASAP (even with Google Postponing sunset) and this feature can be great help!

Hamdan85:
do we have any info on this? Im deploying an app but afraid that it is gonna break on jan 2024

Macil:
Sorry, I was confused by the comments on this issue originally. The InboxSDK is already available with Chrome MV3 support with its npm release. Existing MV2 extensions will need to update from the original "inboxsdk.js" version to using the npm release. Please open a new issue if you have questions or run into problems with that.

Hamdan85:
Is there a documentation for it?

sybers:
Hello,

we have an MV3 Chrome Extension using InboxSDK as well and our beta deployment got rejected yesterday because of a violation in the code they call "Remote hosted code".

I thought the `@inboxsdk/core` package was supposed to work fine with MV3 apps.

If you need more precisions on our setup I'd be happy to provide them :)

bpartridge:
[...]

PR #1086, opened 2023-11-21
handle async thread container element

bkorobeinikov:
Gmail started rendering thread view asynchronously for some users, wait for thread content container element appear in the DOM before notifying with `newGmailThreadView` event

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
@@ -294,6 +294,31 @@ const GmailElementGetter = {
     return document.querySelector(selector_2023_11_16);
   },
 
+  getPreviewPaneThreadContainerElement(): HTMLElement | null | undefined {
+    // This element is always present in thread lists, but it only has contents
+    // when in preview pane mode. We want to monitor it in either case
+    // because the user could switch into preview pane mode.
+    const previewPaneContainer = document.querySelector<HTMLElement>(
+      'div[role=main] .aia',
+    );
+
+    if (!previewPaneContainer) {
+      return null;
+    }
+
+    const selector = 'table.Bs > tr';
+    const selector_2023_11_16 = '.ao8:has(.a98.iY)';
+
+    const threadContainerElement =
+      previewPaneContainer.querySelector<HTMLElement>(selector);
+
+    if (threadContainerElement) {
+      return threadContainerElement;
+    }
+
+    return previewPaneContainer.querySelector<HTMLElement>(selector_2023_11_16);
+  },
+
   getToolbarElement(): HTMLElement {
     return querySelector(document, '[gh=tm]');
   },
--- a/src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
@@ -294,7 +294,7 @@ const GmailElementGetter = {
     return document.querySelector(selector_2023_11_16);
   },
 
-  getPreviewPaneThreadContainerElement(): HTMLElement | null | undefined {
+  getPreviewPaneThreadContainerElement(): HTMLElement | null {
     // This element is always present in thread lists, but it only has contents
     // when in preview pane mode. We want to monitor it in either case
     // because the user could switch into preview pane mode.
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
@@ -406,11 +406,15 @@ class GmailRouteView {
     }
 
     if (parseResult === null) {
+      // check if the page contains list of thread rows
+      const viewWithRows =
+        this.#page.tree.getAllByTag('rowListElement').values().size > 0;
+
       const readingPaneDisabled = !document.querySelector(
         '.bGI[role=main] .Nu.S3.aZ6',
       );
 
-      if (readingPaneDisabled) {
+      if (viewWithRows && readingPaneDisabled) {
         // (.aia) element is not present when reading pane is disabled in Gmail settings
         // avoid logging not found error in this case
       } else {
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
@@ -346,58 +346,9 @@ class GmailRouteView {
   }
 
   async #setupContentAndSidebarView() {
-    let parseResult:
-      | {
-          threadContainerElement?: HTMLElement;
-          previewPaneContainerElement?: HTMLElement;
-        }
-      | 'skip'
-      | null = null;
+    const parseResult = await this.#waitForThreadContainerSafe();
 
-    try {
-      // additionally to page loading state, gmail can render content asynchronously
-      // wait until any of the content container elements appear in the DOM
-      parseResult = await waitFor(() => {
-        if (this.#destroyed) {
-          return 'skip';
-        }
-
-        const threadContainerElement = this._getThreadContainerElement();
-
-        if (threadContainerElement) {
-          return { threadContainerElement };
-        }
-
-        const previewPaneContainerElement =
-          GmailElementGetter.getPreviewPaneContainerElement();
-
-        if (previewPaneContainerElement) {
-          return { previewPaneContainerElement };
-        }
-
-        return null;
-      }, 15_000);
-    } catch {
-      // if user has reading pane disabled in Gmail settings, even preview pane container is not rendered
-      // avoid throwing an error if there are thread rows available
-      const rowsAvailable =
-        this.#page.tree.getAllByTag('rowListElement').values().size > 0;
-
-      if (rowsAvailable) {
-        parseResult = 'skip';
-      } else {
-        const error = new Error("Thread container element wasn't found");
-        if (isStreakAppId(this._driver.getAppId())) {
-          this._driver.getLogger().error(error, {
-            html: extractDocumentHtmlAndCss(),
-          });
-        }
-
-        throw error;
-      }
-    }
-
-    if (parseResult === 'skip' || this.#destroyed) {
+    if (this.#destroyed) {
       return;
     }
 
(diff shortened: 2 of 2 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
```diff
@@ -357,17 +390,12 @@ class GmailRouteView {
...
+    } else if (contentElement?.previewPanelThreadContainerElement) {
+      this._startMonitoringPreviewPaneForThread(
+        contentElement.previewPanelThreadContainerElement,
       );
-
-      if (previewPaneContainer) {
-        this._startMonitoringPreviewPaneForThread(previewPaneContainer);
-      }
+    } else {
+      throw new Error(`Thread content element wasn't found`);
```
wegry, line 398: We expect `threadContainerElement` to come back `null` from `GmailElementGetter` as Gmail loads. `previewPanelThreadContainerElement` must be defined in the first 15s we've switched to a route.

Right? In that case, we should throw a `SelectorError` here still. We've waited 15s for the preview pane to load in, and it hasn't happened at this point.

wegry: How we share that selector information between this file and `GmailElementGetter` is up in the air, but maybe something like this 

```typescript
// selectors.ts
/**
 * To deal with Gmail changes in an MV3 world, we'd like to leave open the door to remote loading selectors as data. @see https://github.com/InboxSDK/InboxSDK/issues/798 for more info.
 */
export const enum Selectors {
  /**
   * This element is always present in thread lists, but it only has contents
   * when in preview pane mode. We want to monitor it in either case
   * because the user could switch into preview pane mode.
   */
  previewPanelThreadContainerElement = 'div[role=main] .aia',
  
  // We might want to work in the GmailElementGetter selectors here as well.  
  // const selector = 'table.Bs > tr';
  // const selector_2023_11_16 = '.ao8:has(.a98.iY)';
}
```

wegry: _This is a bigger ask._ How we share that selector information between this file and `GmailElementGetter` is up in the air, but maybe something like this 

```typescript
// selectors.ts
/**
 * To deal with Gmail changes in an MV3 world, we'd like to leave open the door to remote loading selectors as data. @see https://github.com/InboxSDK/InboxSDK/issues/798 for more info.
 */
export const enum Selectors {
  /**
   * This element is always present in thread lists, but it only has contents
   * when in preview pane mode. We want to monitor it in either case
   * because the user could switch into preview pane mode.
   */
  previewPanelThreadContainerElement = 'div[role=main] .aia',
  
  // We might want to work in the GmailElementGetter selectors here as well.  
  // const selector = 'table.Bs > tr';
  // const selector_2023_11_16 = '.ao8:has(.a98.iY)';
}
```

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
```diff
@@ -392,42 +420,8 @@ class GmailRouteView {
   }
 
   async _startMonitoringPreviewPaneForThread(
```
wegry, line 422: This method isn't called outside this class, and this class isn't part of our public interface.

```suggestion
  async #startMonitoringPreviewPaneForThread(
```

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
```diff
@@ -392,42 +420,8 @@ class GmailRouteView {
...
-        this._driver.getLogger().error(selectorError, {
-          html: extractDocumentHtmlAndCss(),
-        });
-      }
-
-      throw selectorError;
-    }
-
     const elementStream = makeElementChildStream(threadContainerElement).filter(
       (event) =>
```
wegry, line 426: I think we may want to hoist this filter selector logic into `GmailElementGetter` or `selectors`:

```typescript        
        !!event.el.querySelector('.if') ||
        !!event.el.querySelector('.PeIF1d') ||
        !!event.el.querySelector('.a98.iY'),
```

could be rendered as 

```typescript        
        !!event.el.querySelector(':has(.if, .PeIF1d, .a98.iY)'),
```

and might be in a `selectors.ts` type enum as

```typescript
const enum Selectors {
  /**
   * This selector to check the backing element used to construct a ThreadView driver has the correct children.
   */
   threadContainerElementChildIsPresent = ':has(.if, .PeIF1d, .a98.iY)'
}
```

wegry (changes requested):
This PR looks pretty good to me, I think the big things I'd like to see changed are around:

1. Keeping some `SelectorError`s around this codepath for grouping and visibility in Sentry
2. Method visibility and type cruft
3. (stretch) Selector sharing in a way that starts us toward remote loaded config in the vein of #798.

bkorobeinikov:
> 1. Keeping some `SelectorError`s around this codepath for grouping and visibility in Sentry

What would be your proposal for the SelectorError in this case?

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
bkorobeinikov: I don't think this is a good place to start refactoring, this is more like an architectural design decision that needs to be thought through.

wegry:
> > Keeping some SelectorErrors around this codepath for grouping and visibility in Sentry
> 
> What would be your proposal for the SelectorError in this case?

@bkorobeinikov 

I think the preview pane container selector should be the `SelectorError`.

bkorobeinikov:
> > > Keeping some SelectorErrors around this codepath for grouping and visibility in Sentry
> > 
> > 
> > What would be your proposal for the SelectorError in this case?
> 
> @bkorobeinikov
> 
> I think the preview pane container selector should be the `SelectorError` if lookup fails.

still not quite sure, what if the gmail page content is not ready so it needs to wait few seconds until either of the containers appear

wegry:
https://github.com/InboxSDK/InboxSDK/pull/1086/files#r1402020745 is the line I'm thinking we'd have the `SelectorError`.

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
bkorobeinikov: changed

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
bkorobeinikov: yea moving selectors into an enum is a bigger task, would need to be handled as a separate PR

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
```diff
@@ -723,11 +717,13 @@ class GmailRouteView {
...
-      return GmailElementGetter.getThreadContainerElement();
-    },
-  );
+  _getThreadContainerElement(): HTMLElement | null | undefined {
+    return GmailElementGetter.getThreadContainerElement();
+  }
+
+  _getPreviewPaneThreadContainerElement(): HTMLElement | null | undefined {
+    return GmailElementGetter.getPreviewPaneThreadContainerElement();
+  }
```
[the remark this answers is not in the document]
bkorobeinikov: inlined use of `GmailElementGetter.getPreviewPaneThreadContainerElement()`

src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
```diff
@@ -294,6 +294,31 @@ const GmailElementGetter = {
     return document.querySelector(selector_2023_11_16);
   },
 
+  getPreviewPaneThreadContainerElement(): HTMLElement | null | undefined {
```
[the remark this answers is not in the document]
bkorobeinikov: changed

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
```diff
@@ -99,6 +100,8 @@ class GmailRouteView {
   }
 
   destroy() {
+    this.#destroyed = true;
```
wegry, line 103: If we add a `#destroyed` field, does the check [here](https://github.com/InboxSDK/InboxSDK/pull/1086/files#diff-bfcc9a881cfa869791111b1e104c99a8318a2273b4420ec1a74d20b25d4291d4R302) need to change?

I wonder if the `asap` call wrapping an async callback is load bearing, but I doubt it.

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
```diff
@@ -357,18 +364,70 @@ class GmailRouteView {
...
+        }
+
+        const previewPaneContainerElement =
+          GmailElementGetter.getPreviewPaneContainerElement();
+        if (previewPaneContainerElement) {
+          return { previewPaneContainerElement };
+        }
+
+        return null;
+      }, 15_000);
```
wegry, line 403: I wonder if we want to back off how frequently we initially check for these elements. May be related to #1088.

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
```diff
@@ -724,11 +791,9 @@ class GmailRouteView {
       .map((part) => part.substring(1));
   }
 
-  _getThreadContainerElement: () => HTMLElement | null | undefined = once(
-    () => {
-      return GmailElementGetter.getThreadContainerElement();
-    },
-  );
+  _getThreadContainerElement() {
```
wegry, line 794: We should cu this method in its entirety and just use the `GmailElementGetter` equivalent directly.

Macil (approved):

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
bkorobeinikov: ill keep existing method because it is being used in few more places, want to limit amount of changes for this fix

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
bkorobeinikov: adding a check for `#destroyed` is not totally required in this case. Will keep this part unchanged to minimize the diff.

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
bkorobeinikov: hm might be related, can't find a codepath that would throw exception with selector `[role=main] .aia`

PR #1087, opened 2023-11-22
proposed fix for: SelectorError: Failed to find element with selector: table.Bs > tr, .ao8:has(.a98.iY)

### LINKED ISSUE #1078: `Compose.registerComposeViewHandler` Registered Compose View Handler does not fire on Reply/Reply All/Forward

Alex-Chopard:
Hi, sadly we experiencing more issues with the preview pane and the detection of a selected conversation / switch to another conversation.
As I'm able to reproduce the issue on the test account, I trying a fix.

Fill free to ask if you have any question / suggestion about this PR.

Error in the console:
Screenshot 2023-11-22 at 11 58 46

DOM with preview panel enabled and no conversation selected:
`.ao8` is well found but not `.a98.iY`
Screenshot 2023-11-22 at 12 00 33

DOM with preview panel enabled and a conversation selected:
`.apa` has been replaced with `.a98.iY`
Screenshot 2023-11-22 at 12 01 28

Without the fix:

https://github.com/InboxSDK/InboxSDK/assets/29271954/33666754-2b65-40d0-b863-8b736a854ad0

With the fix:

https://github.com/InboxSDK/InboxSDK/assets/29271954/3c2363a8-d635-4cf6-9e62-83f51d3bb3eb

#1071

Closes #1078

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
@@ -433,8 +433,7 @@ class GmailRouteView {
         !!event.el.querySelector('.if') ||
         !!event.el.querySelector('.PeIF1d') ||
         !!event.el.querySelector('.a98.iY') ||
-        (event.el.classList.contains('a98') &&
-          event.el.classList.contains('iY')),
+        event.el.matches('.a98.iY'),
     );
 
     this._eventStream.plug(
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
@@ -397,7 +397,7 @@ class GmailRouteView {
     let threadContainerElement;
 
     const selector = 'table.Bs > tr';
-    const selector_2023_11_16 = '.ao8:has(.a98.iY)';
+    const selector_2023_11_16 = '.ao8:has(.a98.iY), .ao9:has(.apa)';
 
     try {
       threadContainerElement = await waitFor(() => {
@@ -432,7 +432,9 @@ class GmailRouteView {
       (event) =>
         !!event.el.querySelector('.if') ||
         !!event.el.querySelector('.PeIF1d') ||
-        !!event.el.querySelector('.a98.iY'),
+        !!event.el.querySelector('.a98.iY') ||
+        (event.el.classList.contains('a98') &&
+          event.el.classList.contains('iY')),
     );
 
     this._eventStream.plug(


src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
```diff
@@ -432,7 +432,9 @@ class GmailRouteView {
       (event) =>
         !!event.el.querySelector('.if') ||
         !!event.el.querySelector('.PeIF1d') ||
-        !!event.el.querySelector('.a98.iY'),
+        !!event.el.querySelector('.a98.iY') ||
+        (event.el.classList.contains('a98') &&
+          event.el.classList.contains('iY')),
```
wegry, line 437: nit: Could this be simplified to

```typescript
        event.el.matches('.a98.iY'),
```

wegry:
Thanks for this @Alex-Chopard. I'll try testing against my accounts to see if the existing behavior holds there. @bkorobeinikov might want to take a look as well. We'll try getting back to you by 18:00 UTC about pulling this in.

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
wegry, line 437: nit: Could this be simplified to

```suggestion
        !!event.el.querySelector('.a98.iY') ||
        event.el.matches('.a98.iY'),
```

Alex-Chopard: Indeed, I did not know this method, thanks 🙏

wegry (approved):
I'm inclined to say we should move forward with this PR. It may fix both #1071 and #1078. I would like another one of the folks I've flagged as a reviewer bless it though.

wegry:
This is included in `@inboxsdk/core@2.1.19`.

PR #1089, opened 2023-11-27
Fix issue with typings related to the definition of the `PublicOnly` type

aleksandr-lapushkin:
### Overview
Currently, running `gulp` or `gulp types` will output an incorrect file structure. This is present in the published library starting with `2.1.13`. Specifically, the `src` directory gets included twice:
image

As a result, the typings in `inboxsdk.d.ts` are broken, since the paths are no longer correct:
image

This is a regression introduced in `2.1.13` with this change: https://github.com/InboxSDK/InboxSDK/commit/50b583bf076193e3cba3b985324765bdb5c61bf1#diff-b5c5b3d03297cce43ffb834917bfdc3d10fc8aefa37979c5af23c2205d1bc88cR1-R9

### Proposed changes
This PR moves the type in question into the project source files, thereby fixing the above issue.

### Steps to reproduce
Steps to reproduce on `main`:
1. `yarn`
2. `yarn gulp`


aleksandr-lapushkin:
@wegry Hey, would appreciate if you could take a look at this. Thanks!

PR #1095, opened 2023-11-29
fix selector of preview pane opened thread view

bkorobeinikov:
`.ao9:has(.a98.iY)` selector for preview pane with opened email.

if user navigates to inbox page with preview pane enabled, opens email, navigates to another page, and then navigates back to inbox, gmail will pre open email from earlier. Having .ao8 as a thread container was preventing MutationObserver identifiying changes to the thread view dom, moving the thread container down one node (to .ao9) allows MutationObserver to identify the changes when switching between emails.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
@@ -460,7 +460,7 @@ class GmailRouteView {
     let threadContainerElement: HTMLElement | 'destroyed';
 
     const selector = 'table.Bs > tr';
-    const selector_2023_11_16 = '.ao8:has(.a98.iY), .ao9:has(.apa)';
+    const selector_2023_11_16 = '.ao9:has(.a98.iY), .ao9:has(.apa)';
 
     try {
       threadContainerElement = await waitFor(() => {


src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
```diff
@@ -460,7 +460,7 @@ class GmailRouteView {
     let threadContainerElement: HTMLElement | 'destroyed';
 
     const selector = 'table.Bs > tr';
-    const selector_2023_11_16 = '.ao8:has(.a98.iY), .ao9:has(.apa)';
+    const selector_2023_11_16 = '.ao9:has(.a98.iY), .ao9:has(.apa)';
```
wegry, line 463:
```suggestion
    const selector_2023_11_16 = '.ao9:has(.a98.iY, .apa)';
```

We may want to update the date in the variable name.

wegry:
We may want to remove the check here: https://github.com/InboxSDK/InboxSDK/blob/3881b632cf22437db82bceedafefbfa21f2f737e/src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts#L507 as well.

wegry:
Related to #1087.

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
bkorobeinikov: it is same date, choosing .ao8 was a mistake (altho we couldn't repro ourselves so we didn't know about mutation observer issue)

bkorobeinikov:
> We may want to remove the check here:

nope that check is needed

src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
bkorobeinikov: the suggested change is not much different from existing code, will keep as is

PR #1096, opened 2023-11-30
chore(ThreadView): log selector existence

wegry:
We have a number of selectors around ThreadViews now. This PR adds logging for which selector versions are actually in use so we can potentially simplify logic in the future.


wegry:
Prelim data shows that only the [`2023-11-22` selector](https://github.com/InboxSDK/InboxSDK/pull/1096/files#diff-bfcc9a881cfa869791111b1e104c99a8318a2273b4420ec1a74d20b25d4291d4R515) added in #1087 is still used.

[Streak internal link](https://streak.slack.com/archives/C02J7RSAUJ0/p1701356517499679) about checking usage

PR #1118, opened 2024-01-29
feat(SectionView): update title and subtitle support

wegry:
It may be simpler to just update to the newer styles without configurability. We can always add classname configuration later.

# Screenshots

### Before
<figure>
Screenshot 2024-02-16 at 09 21 34
<figcaption>Note the hyphen between 'Inbox Monkeys' and 'Subtitle'. Also note the parens around 'Subtitle'.</figcaption>
</figure>

### After

Screenshot 2024-02-16 at 09 22 02

### Before

Screenshot 2024-02-21 at 14 52 14

### After

Screenshot 2024-02-21 at 14 26 46

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-collapsible-section-view.module.css
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-collapsible-section-view.module.css
@@ -1,0 +1,4 @@
+.subtitle {
+  margin-left: 1em;
+  color: rgb(94 94 94);
+}
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-collapsible-section-view.module.css
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-collapsible-section-view.module.css
@@ -1,11 +1,11 @@
 :global(.Wn) {
-  --inboxsdk-title-color: rgb(0 0 0 / 54%);
-  --inboxsdk-subtitle-color: rgb(0 0 0 / 87%);
+  --inboxsdk-title-color: rgb(0 0 0 / 87%);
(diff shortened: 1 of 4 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-collapsible-section-view.module.css
```diff
@@ -0,0 +1,16 @@
+.title {
+  color: rgb(0 0 0 / 54%) !important;
+}
+
+.subtitle {
+  margin: 0;
+  color: rgb(0 0 0 / 87%) !important;
```
ruthhalteman, line 7: is there a way to use specificity instead of `!important` here? Seems like it might cause the user to have a worse time if they have to use `!important` and/or not have their styling work?

wegry: The trickiness here is that for title and subtitle, Gmail has `.Wr>.Wn` as the ruleset selector setting color. For the CSS local `.subtitle`, specifically, we could do something like `.title .subtitle`. For `.title` though, we would need to pick a parent.

wegry: Yup, we've used the same title parent selector for 5 years though, so maybe that's safe as a parent to use here too.

wegry: See [ac037be](https://github.com/InboxSDK/InboxSDK/pull/1118/commits/ac037be673306d5d8891c08f714910e02e41fa94).

ruthhalteman (approved):

PR #1121, opened 2024-02-07
Being able to add buttons to the right of the subject line in thread view

ruthhalteman:
Streak example is our streak share button. Next to gmail's print and open in new buttons.

Currently draft because I don't know how correct or robust this is. Will keep testing. So far the main issue I see is that it's not finding the container in the reading pane views. Not sure why, the subject line label (existing functionality) is fine and works in a very similar way.

Summary of additions:
- Thread view function `addSubjectButton` that takes in a button descriptor and makes a button using our Button view
- Mutation observer for cases when the subject button container is lazy loaded
- Fix for case when a `ButtonView` doesn't have a dropdown where the focus class doesn't get removed

[Relevant Streak box](https://www.streak.com/a/boxes/agxzfm1haWxmb29nYWVyNAsSDE9yZ2FuaXphdGlvbiIRb2lzbWFpbEBnbWFpbC5jb20MCxIEQ2FzZRiAgO7r38zdCQw)

Known issues:
- [x] Spacing between SDK buttons isn't good/doesn't match gmail buttons (when adding two or more)


src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
```diff
@@ -625,6 +627,25 @@ class GmailThreadView {
...
   }
 
+  addSubjectButton(button: any) {
+    const subjectToolbarElement = this._findSubjectToolbarElement();
+    if (!subjectToolbarElement) {
+      throw new Error('Subject toolbar element not found');
+    }
+    // gmail has these spacer spans. Not sure if important or a good idea to use this way
+    const el = document.createElement('span');
+    el.setAttribute('data-unique-tt-id', 'ucc-2');
```
ruthhalteman, line 637: oh yeah and it shouldn't be hardcoded to 2 if this is actually kept in here. should be consecutive like the gmail ones are

src/platform-implementation-js/views/conversations/thread-view.ts
```diff
@@ -143,6 +143,10 @@ class ThreadView extends (EventEmitter as new () => TypedEventEmitter<ThreadView
   addLabel(): SimpleElementView {
     return get(memberMap, this).threadViewImplementation.addLabel();
   }
+
+  addSubjectButton(buttonDescriptor: Record<string, any>) {
```
wegry, line 147: Could we use https://github.com/InboxSDK/InboxSDK/blob/4edf7ccd5e69417933291d69006d111f7bfc4bf7/src/inboxsdk.d.ts#L446 here?

```suggestion
  addSubjectButton(buttonDescriptor: Descriptor<ButtonDescriptor>) {
```

src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
```diff
@@ -687,6 +708,12 @@ class GmailThreadView {
     return null;
   }
 
+  _findSubjectToolbarElement(): HTMLElement | null | undefined {
```
wegry, line 711: Unless we want this method visible outside the class...

```suggestion
  #findSubjectToolbarElement(): HTMLElement | null {
```

src/platform-implementation-js/views/conversations/thread-view.ts
wegry: If we don't necessarily want an `Observable` as part of the API, an alternative would be 

```suggestion
  addSubjectButton(buttonDescriptor: ButtonDescriptor) {
```

src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
```diff
@@ -625,6 +627,25 @@ class GmailThreadView {
     return view;
   }
 
+  addSubjectButton(button: any) {
```
wegry, line 630: We might use either `ButtonDescriptor` or `Descriptor<ButtonDescriptor>` as the type here.

wegry (commented):
Looks pretty good so far.

I did notice the subject button disappearing after 5s or so with the reading pane enabled...

See https://www.loom.com/share/0dc58ca527294d7598dffb813600ae6d

src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
```diff
@@ -687,6 +737,12 @@ class GmailThreadView {
     return null;
   }
 
+  _findSubjectToolbarElement(): HTMLElement | null {
+    var toolbarContainerElements =
+      this._element.querySelectorAll<HTMLElement>('.bHJ');
+    return toolbarContainerElements[0];
```
wegry, line 743: What's the rationale for querySelectorAll here?

Maybe we just want:

```suggestion
    const toolbarContainerElement =
       this._element.querySelector<HTMLElement>('.bHJ');
     return toolbarContainerElement;
```

wegry (changes requested):
Alright, so running through this example extension with the reading pane enabled and not looks good.

Few things I'd like to see changed prior to pulling it in are (in order):

1. Use `SelectorError` for reporting's sake when an element isn't present and we expect it to be so.
1. Plumb the ButtonDescriptor all the way through from the public interface (removing some remaining `any`s).
2. Not adding new non public methods to gmail-thread-view).

wegry (approved):

src/platform-implementation-js/views/conversations/thread-view.ts
```diff
@@ -164,6 +164,10 @@ class ThreadView extends (EventEmitter as new () => TypedEventEmitter<ThreadView
         });
       });
   }
+
```
wegry, line 167:
```suggestion
  /**
   * @internal
   */
```

PR #1124, opened 2024-02-14
fix(ComposeView): clean up dropdown to prevent empty dropdown from toggling

wegry:
This particular bug triggers when a ComposeView.addButton is used with a descriptor. If the descriptor goes from hasDropdown === true to false, we currently trigger an empty dropdown when the button is subsequently clicked.

[Loom](https://www.loom.com/share/6a1fbcb560aa43088b3da5957bbfc149)

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/add-button.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/add-button.ts
@@ -9,7 +9,6 @@ import GmailDropdownView from '../../widgets/gmail-dropdown-view';
 import insertElementInOrder from '../../../../lib/dom/insert-element-in-order';
 import type GmailComposeView from '../gmail-compose-view';
 import type { ComposeButtonDescriptor } from '../../../../driver-interfaces/compose-view-driver';
-import type { Options } from '../../../../views/compose-button-view';
 import type { Bus } from 'kefir-bus';
 
 export type AddedButtonEvents = {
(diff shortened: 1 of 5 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/add-button.ts
```diff
@@ -34,6 +34,14 @@ export default function addButton(
           gmailComposeView.getGmailDriver(),
         );
 
+        if (
+          buttonViewController instanceof DropdownButtonViewController &&
+          buttonOptions?.hasDropdown !== true
+        ) {
+          buttonViewController.destroy();
+          buttonViewController = undefined;
```
wegry, line 42: Because we call `resolve` below this on [line 52](https://github.com/InboxSDK/InboxSDK/pull/1124/files#diff-0cc9418b0cfda04a71c671ef1a2be0fc8bcdaa878ff68eeae10e7820c97b5089L44), this approach might not be workable.

`ComposeButtonView` a few frames higher is only using the initial `Promise.resolve`. Any subsequent `resolve` call after the first is a no-op.

src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/add-button.ts
```diff
@@ -19,6 +28,7 @@ export default function addButton(
   >,
   groupOrderHint: string,
   extraOnClickOptions: Record<string, any>,
+  bus: Bus<AddedButtonEvents, unknown>,
 ) {
   return new Promise<Options | null>((resolve) => {
```
wegry, line 33: We should probably remove the wrapping Promise here to account for the new `Bus` API.

PR #1125, opened 2024-02-14
chore: add logging for suggestionsResults for streak users

wegry:
We're seeing what might be Gmail rolling out a change to its search results format. Logging example requests and responses here will help narrow down if that's the case.

### DIFF
--- a/src/injected-js/gmail/setup-gmail-interceptor.ts
+++ b/src/injected-js/gmail/setup-gmail-interceptor.ts
@@ -719,17 +719,7 @@ export function setupGmailInterceptorOnFrames(
           const modifications = await (connection as any)._defer.promise;
 
           if (modifications) {
-            const modified = modifySuggestions(responseText, modifications);
-            logger.eventSdkPassive(
-              'suggestionsModified',
-              {
-                query: currentQuery,
-                originalResponseText: responseText,
-                modifiedResponseText: modified,
-              },
-              true,
-            );
-            return modified;
+            return modifySuggestions(responseText, modifications);
           }
         }
 
--- a/src/injected-js/gmail/setup-gmail-interceptor.ts
+++ b/src/injected-js/gmail/setup-gmail-interceptor.ts
@@ -719,10 +719,28 @@ export function setupGmailInterceptorOnFrames(
           const modifications = await (connection as any)._defer.promise;
 
           if (modifications) {
-            return modifySuggestions(responseText, modifications);
+            const modified = modifySuggestions(responseText, modifications);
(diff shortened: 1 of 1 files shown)


src/injected-js/gmail/setup-gmail-interceptor.ts
```diff
@@ -719,10 +719,28 @@ export function setupGmailInterceptorOnFrames(
...
+            logger.eventSdkPassive(
+              'suggestionsModified',
+              {
+                query: currentQuery,
+                originalResponseText: responseText,
+                modifiedResponseText: modified,
+              },
+              true,
+            );
+            return modified;
```
jfmaggie, line 732: we are gonna have a lot of these logged and i am not sure if it's worth it. we could check the modified part running in dev and only log the failed part like below


```suggestion
            return modifySuggestions(responseText, modifications);
```

jfmaggie (approved):
lgtm

src/injected-js/gmail/setup-gmail-interceptor.ts
wegry: > we are gonna have a lot of these logged and i am not sure if it's worth it.

It looks like this if block is hit when modifications successfully come back or a request finishes.

> we could check the modified part running in dev and only log the failed part like below.

Based on local logging, it looks like with this suggestion running in prod, we're logging all keystrokes but ones that have suggestions responses come back. I think we might want to flip the logging logic to _only_ log when suggestions have been modified or if this function throws for whatever reason.

wegry: If you'd like to see how modifySuggestions works, here's how I tested.

```diff
diff --git a/src/injected-js/gmail/setup-gmail-interceptor.ts b/src/injected-js/gmail/setup-gmail-interceptor.ts
index f63b121f..9ce1b093 100644
--- a/src/injected-js/gmail/setup-gmail-interceptor.ts
+++ b/src/injected-js/gmail/setup-gmail-interceptor.ts
@@ -716,7 +716,14 @@ export function setupGmailInterceptorOnFrames(

       async responseTextChanger(connection, responseText) {
         if ((connection as any)._defer && connection.status === 200) {
-          const modifications = await (connection as any)._defer.promise;
+          let modifications;
+
+          try {
+            modifications = await (connection as any)._defer.promise;
+            console.log('modifications promise succeeded', modifications);
+          } catch (e) {
+            console.log('modifications promise threw', e);
+          }

           if (modifications) {
             return modifySuggestions(responseText, modifications);
```

PR #1126, opened 2024-02-14
fix: use trustedTypes and log when modifySuggestions throws an error

wegry:
Add trustedTypes with Safari fallback to account for folks starting to get https://workspaceupdates.googleblog.com/2024/01/extending-trusted-types-to-gmail.html as part of Rapid Release.

And log errors on modify suggestions instead of when a suggestions request comes back with `undefined` `modifications`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "combokeys-capture": "^2.4.2",
     "concat-stream": "^2.0.0",
     "contain-by-screen": "^1.0.3",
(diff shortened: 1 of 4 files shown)


jfmaggie:
can you verify example extension Search Suggestions still works pls?

wegry:
> can you verify example extension Search Suggestions still works pls?

@jfmaggie checked against the Streak extension proper. I'll check against search suggestions in the morning.

PR #1137, opened 2024-02-26
Make html-to-text work without dompurify

Macil:
Follow-up to https://github.com/InboxSDK/InboxSDK/pull/1126. Removes dompurify which was more heavyweight and unnecessary here.

No visible behavior changes.


wegry (approved):
LGTM. Thanks for the changes.

PR #1140, opened 2024-02-28
types: typed events for section views

wegry:
Prep for layout work around `CollapsibleSectionView`.

Runs types through events for `SectionView` and `CollapsibleSectionView`. Requires a bit of runtime rejiggering to narrow event types.

tip: ignore whitespace.

### DIFF
--- a/src/platform-implementation-js/views/section-view.ts
+++ b/src/platform-implementation-js/views/section-view.ts
@@ -1,13 +1,15 @@
 import EventEmitter from '../lib/safe-event-emitter';
 import type { Driver } from '../driver-interfaces/driver';
 import type GmailCollapsibleSectionView from '../dom-driver/gmail/views/gmail-collapsible-section-view';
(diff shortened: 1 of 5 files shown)


src/platform-implementation-js/views/section-view.ts
```diff
@@ -37,46 +39,50 @@ function _bindToEventStream(
...
-        }
-
-        if (typeof rowDescriptor.onClick === 'function') {
-          rowDescriptor.onClick();
-        }
-      },
-    );
-  sectionViewDriver
-    .getEventStream()
-    .filter(({ eventName }) => eventName === 'summaryClicked')
```
wegry, line 65: This looks like it was mapped to `titleLinkClicked` in collapsible-section-view back in d38e019f, but not in this file. We don't depend on it downstream at Streak, but looks like an old bug. 

Otherwise, all event names should match between old and new code.

PR #1145, opened 2024-03-03
Filter duplicate IDs to ensure custom lists with attachments will always render

Roensby:
…ays render

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/gmail-driver/show-custom-thread-list.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/gmail-driver/show-custom-thread-list.ts
@@ -1,4 +1,5 @@
 import find from 'lodash/find';
+import uniqBy from 'lodash/uniqBy';
 import GmailElementGetter from '../gmail-element-getter';
 import * as SyncGRP from '../gmail-sync-response-processor';
 import type Logger from '../../../lib/logger';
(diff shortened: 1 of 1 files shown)


Roensby:
hm, looks like these tests fail for other reasons?

src/platform-implementation-js/dom-driver/gmail/gmail-driver/show-custom-thread-list.ts
```diff
@@ -249,7 +249,16 @@ const setupSearchReplacing = (
           threadDescriptors = [];
         }
 
+        const visited = new Set<InitialPair>();
```
wegry, line 252:
```suggestion
        const visited = new Set<InitialIDPair>();
```

Should patch up tests.

wegry: You'll need to run `tsc` locally to patch up the types here.

tsc output as of [3b76776](https://github.com/InboxSDK/InboxSDK/pull/1145/commits/3b76776141fb92d13ac0559f01dff3cfa6d09ad2)

```
src/platform-implementation-js/dom-driver/gmail/gmail-driver/show-custom-thread-list.ts:255:29 - error TS2345: Argument of type 'ThreadDescriptor' is not assignable to parameter of type 'InitialIDPair'.
  Type 'string' is not assignable to type 'InitialIDPair'.

255             if (visited.has(threadDescriptor)) {
                                ~~~~~~~~~~~~~~~~

src/platform-implementation-js/dom-driver/gmail/gmail-driver/show-custom-thread-list.ts:259:25 - error TS2345: Argument of type 'ThreadDescriptor' is not assignable to parameter of type 'InitialIDPair'.

259             visited.add(threadDescriptor);
                            ~~~~~~~~~~~~~~~~


Found 2 errors in the same file, starting at: src/platform-implementation-js/dom-driver/gmail/gmail-driver/show-custom-thread-list.ts:255


Exited with code exit status 2
```

Macil:
I've tweaked the PR so it does the duplicate check at a later stage after we've converted any message ids into thread ids which could also create more duplicates.

Thanks for investigating this. We'll get this fix out in the next version.

PR #1146, opened 2024-03-04
feat: allow passing arbitrary HTML to ThreadRow addImage

wegry:
Uses same interface as #1133.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.ts
@@ -21,7 +21,9 @@ import GmailLabelView from '../widgets/gmail-label-view';
 import GmailActionButtonView from '../widgets/gmail-action-button-view';
 import type GmailDriver from '../gmail-driver';
 import type GmailRowListView from './gmail-row-list-view';
-import updateIcon from '../../../driver-common/update-icon';
+import updateIcon, {
+  type IconSettings,
+} from '../../../driver-common/update-icon';
 import type {
   Contact,
   ImageDescriptor,
@@ -29,6 +31,7 @@ import type {
   ThreadDateDescriptor,
 } from '../../../../inboxsdk';
 import { assert } from '../../../../common/assert';
+import type { Descriptor } from '../../../../types/descriptor';
 
 type LabelMod = {
   gmailLabelView: Record<string, any>;
@@ -44,7 +47,7 @@ type ButtonMod = {
   remove(): void;
 };
 type ImageMod = {
-  iconSettings: Record<string, any>;
+  iconSettings: IconSettings;
   iconWrapper: HTMLElement;
   remove(): void;
 };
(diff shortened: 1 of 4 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.ts
```diff
@@ -430,28 +437,28 @@ class GmailThreadRowView {
...
     if (!this._elements.length) {
       console.warn('addImage called on destroyed thread row');
       return;
     }
 
-    const prop = kefirCast(Kefir, inIconDescriptor)
+    const prop: Kefir.Observable<[ImageDescriptor | null], unknown> = kefirCast(
+      Kefir,
+      inIconDescriptor,
+    )
```
bkorobeinikov, line 449: to avoid any casting, we can cast only the kefirCast result

```
kefirCast(Kefir, inIconDescriptor) as Kefir.Observable<ImageDescriptor | null, unknown>
```

and then add a second paramter to `combine` to define the combinator explicitly
```
.combine(
    Kefir.merge(....),
    (iconDescriptor, _) => [iconDescriptor],
);
```

src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.ts
```diff
@@ -463,52 +470,74 @@ class GmailThreadRowView {
...
+
+      updateIcon(
+        iconSettings,
+        iconWrapper,
+        false,
+        iconDescriptor.imageClass,
+        iconDescriptor.imageUrl,
+      );
+
+      if (iconDescriptor.image) {
```
bkorobeinikov, line 508: the `unmount` should happen when imageMod gets removed. 

will this be a good option?

```
if (iconDescriptor.image) {
    const child = iconWrapper.appendChild(...);
    const dispose = iconDescriptor.image({el});

    imageMod.disposeImage = dispose;
}
```

the iconDesciptor will return as a result a callback to cleanup. This is a common patter for subscription objects or similar to react useEffect.

then where imageMod is being created, add support for the disposeImage method

```
imageMod = {
            ...
            disposeImage: null,

            remove() {
              this.disposeImage?.();
              this.iconWrapper.remove();
            },
          };
```

src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.ts
```diff
@@ -463,52 +470,74 @@ class GmailThreadRowView {
...
-        const containerRow =
-          this._elements.length === 3 ? this._elements[2] : this._elements[0];
-        containerRow.classList.add('inboxsdk__thread_row_image_added');
 
-        if (iconDescriptor.tooltip) {
-          iconWrapper.setAttribute('data-tooltip', iconDescriptor.tooltip);
-        } else {
-          iconWrapper.removeAttribute('data-tooltip');
-        }
+        const child = iconWrapper.appendChild(document.createElement('div'));
```
bkorobeinikov, line 515: should the `child` be created on each stream event? Looks like in this code a lot of checks are added to prevent duplicated creation of iconWrapper, so i think for the image element it should also only created it during imageMod creation for example.

bkorobeinikov (commented):
left suggestion on type safety, unmounting, and concerns around handling events and element creation

src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.ts
wegry: This is indeed leaking memory. @Macil pointed out elsewhere that due to optimizations around SDK routing, we leave elements in the DOM.

wegry:
https://github.com/InboxSDK/InboxSDK/pull/1146#discussion_r1511696460

PR #1147, opened 2024-03-04
ThreadRowView.addImageGroup() method

Macil:
Alternative to https://github.com/InboxSDK/InboxSDK/pull/1146


src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.ts
```diff
@@ -48,6 +49,11 @@ type ImageMod = {
   iconWrapper: HTMLElement;
   remove(): void;
 };
+interface GroupedImageMod {
+  groupElement: HTMLElement;
+  groupIconSettings: Array<Record<string, any>>;
```
wegry, line 55: Use `ImageDescriptor` here and in `ImageMod`.

Macil:
Shelving this for now, may revisit when we add better support for avatar stacks and overflow menu on thread row.

PR #1150, opened 2024-03-06
ThreadRowView.addImage onHover callback

Macil:
Adds an onHover callback to ThreadRowView.addImage's ImageDescriptor. The callback gives an object with a `hoverEnd: Promise<void>;` property that resolves when the hover finishes.

## Suggested Usage

Intended for event logging within Streak to see how often the addImage tooltip is used.
Highly recommend that for this use, only hovers that last for a certain amount of time like a second are logged in order to only see intentional hovers. This can be done like the following code examples, one using Kefir and one not:

```ts
threadRowView.addImage({
  // ...
  onHover(event) {
    Kefir.later(1000)
      .takeUntilBy(Kefir.fromPromise(event.hoverEnd))
      .onValue(() => {
        logHover();
      });
  }
});
```

```ts
threadRowView.addImage({
  // ...
  onHover(event) {
    const timeout = setTimeout(() => {
      logHover();
    }, 1000);
    event.hoverEnd.then(() => {
      clearTimeout(timeout);
    });
  }
});
```

### DIFF
--- a/src/platform-implementation-js/views/thread-row-view.ts
+++ b/src/platform-implementation-js/views/thread-row-view.ts
@@ -15,6 +15,7 @@ export interface ImageDescriptor {
   imageClass?: string;
   tooltip?: string;
   orderHint?: number;
+  onHover?: (event: { hoverEnd: Promise<void> }) => void;
 }
 
 type EmitterType = TypedEventEmitter<{ destroy: () => void }>;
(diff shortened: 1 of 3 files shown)


src/platform-implementation-js/views/thread-row-view.ts
```diff
@@ -15,6 +15,7 @@ export interface ImageDescriptor {
   imageClass?: string;
   tooltip?: string;
   orderHint?: number;
+  onHover?: (event: { hoverEnd: Promise<void> }) => void;
```
wegry, line 18:
```suggestion
  /** @internal */
  onHover?: (event: { hoverEnd: Promise<void> }) => void;
```

wegry (approved):

src/platform-implementation-js/views/thread-row-view.ts
Macil: I think it's fine to expose it regularly. It's a simple API that doesn't need to change much and if we had to drop support for it, it wouldn't necessarily break applications using the live-loaded version.

PR #1154, opened 2024-03-23
Add Firefox support

### LINKED ISSUE #1105: FR: Firefox Support
What needs to be done for Firefox Compatibility? I'm happy to help if you let me know where to start.

DaPotatoMan:
resolves #1105
This PR adds support for Firefox.

#### Background script
```ts
import '@inboxsdk/core/background'
```

#### Content script
```ts
import '@inboxsdk/core/firefox'
import InboxSDK from '@inboxsdk/core'

InboxSDK.load(...);
```

<details>
<summary>
<h3>Additional Notes</h3>
</summary>

```ts
import '@inboxsdk/core/firefox'
```

This additional import is mandatory until firefox fixes MAIN world execution in content scripts ([See here](https://bugzilla.mozilla.org/show_bug.cgi?id=1736575)).
</details>


wegry:
It's my understanding our MV2 support for Chrome isn't long for this world. If we intend to merge this in, I believe we'd probably only do it for MV3.

DaPotatoMan:
@wegry 
Hi! The changes in this PR aren't specific to MV2 or MV3. It will allow Firefox devs to manually inject `pageWorld.js` until this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1736575) gets fixed. After that InboxSDK will be able to inject the script normally.

Also, Mozilla has decided to keep supporting MV2 for the foreseeable future [(See here)](https://blog.mozilla.org/addons/2024/03/13/manifest-v3-manifest-v2-march-2024-update/). So, the above pattern can be used to support InboxSDK in firefox MV2 without any changes from InboxSDK side.

examples/firefox-mv2/content.js
```diff
@@ -0,0 +1,34 @@
...
+      title: 'My button',
+      iconUrl: (globalThis.chrome || globalThis.browser).runtime.getURL(
+        'monkey.png',
+      ),
+      onClick() {
+        alert('my button works!');
+      },
+    });
+  });
+});
```
wegry, line 34: After touching base with @Macil about this, I think we're in agreement that we _wouldn't_ be willing to accept new work including an MV2 version.

```suggestion
```

examples/firefox-mv2/manifest.json
```diff
@@ -0,0 +1,18 @@
...
+  },
+  "content_scripts": [
+    {
+      "matches": ["*://mail.google.com/*"],
+      "js": ["inboxsdk.js", "content.js"],
+      "run_at": "document_end"
+    }
+  ],
+  "web_accessible_resources": ["pageWorld.js", "monkey.png"]
+}
```
wegry, line 18:
```suggestion
```

examples/firefox-mv2/monkey.png
wegry, line 1: We should remove this file.

src/common/get-extension-id.ts
```diff
@@ -1,7 +1,11 @@
...
+
 export default function getExtensionId(): string | null {
-  const chrome: any = (global as any).chrome;
-  if (chrome && chrome.extension && chrome.extension.getURL) {
-    return chrome.extension.getURL('');
+  try {
+    return browser.runtime.id;
+  } catch (error) {
+    console.error('Failed to get extension id');
   }
```
wegry, line 8: I'm a little trepidatious about `webextension-polyfill` would break our Safari support. Specifically, [this issue](https://github.com/mozilla/webextension-polyfill/issues/364).

This draft block will trigger a `console.error` for Safari extensions using the SDK. 

```suggestion
  return chrome?.extension?.getURL?.('') ?? browser?.runtime?.id;
```

src/common/load-script.ts
wegry, line 1: This file should be reverted as it currently only deals with MV2.

wegry (changes requested):
@Macil and I would want any new MV2 specific bits to be dropped prior to accepting this PR.

DaPotatoMan:
@wegry After thinking it over, I've removed `webextension-polyfill` since InboxSDK only uses extension code that are similar in both Firefox and Chrome (mv3).

`browser` variable is now exported from here:
https://github.com/DaPotatoMan/InboxSDK/blob/c3a8534418d6d713337b84897b6e687e4e3be7b0/src/common/extension-apis.ts#L1-L2

src/common/get-extension-id.ts
DaPotatoMan: Since I've removed webextension-polyfill the code now uses `browser.runtime.id` (chrome.runtime.id).
Also `chrome.extension.getURL` is deprecated in MV3 [(See here)](https://developer.chrome.com/docs/extensions/reference/api/extension#method-getURL).

wegry (approved):
Thanks for the changes.

src/common/extension-apis.ts
```diff
@@ -0,0 +1,14 @@
+/** Refers to {@link chrome} or `browser` in firefox */
+export const browser = globalThis.chrome || (globalThis as any).browser;
+
+if (!browser) {
+  throw new Error('chrome or browser not available in current context.');
+}
+
+export function getExtensionId() {
+  try {
+    return browser.runtime.id;
```
wegry, line 10: This appears to work in Safari Web Extensions. Is there a browser this would throw in? https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/id has Chrome, Edge, FF, and Safari all supporting it in various forms.

Macil:
I don't think we should expose `InboxSDK.setInjectScriptImplementation()` and make it the user's responsibility like this. I'd rather that the injectScriptImplementation inside src/platform-implementation-js/lib/inject-script.ts checked if the browser was Firefox (maybe just through user-agent sniffing with `/firefox/i.test(navigator.userAgent)` unless there's a better way) and injected the script this way, or maybe have a separate import for Firefox support like `import InboxSDK from "@inboxsdk/core/firefox";` (which just did something like `import InboxSDK from "."; InboxSDK.setFirefox(); export default InboxSDK;`).

wegry:
> maybe have a separate import for Firefox support like import InboxSDK from "@inboxsdk/core/firefox"; (which just did something like import InboxSDK from "."; InboxSDK.setFirefox(); export default InboxSDK;)

@Macil I'd lean towards this approach over the UA sniffing. You might recall that the way ESM statically evaluates imports and exports, we've had to do a bit more lifting to get that approach to work. [loadScript](https://github.com/DaPotatoMan/InboxSDK/blob/c3a8534418d6d713337b84897b6e687e4e3be7b0/src/inboxsdk-js/loading/platform-implementation-loader.ts#L13) is hung off a class as opposed to being module level for that reason.

Macil:
The injectScriptImplementation() function inside src/platform-implementation-js/lib/inject-script.ts already has a path where it falls back to loading the script through adding a script tag to the page, so it would be a small change without re-architecting anything to add another case that it does so.

Macil:
@wegry brought to my intention that the injectScriptImplementation() function inside `src/platform-implementation-js/lib/inject-script.ts` is set up to avoid including any references to direct script injection so we don't fail some certain Chrome Web Store static analysis of our npm/MV3 build, so we can't put an if-check in there for Firefox to do a direct script injection as I first suggested above.

Spelling out my second suggestion a little further: we can add a packages/core/firefox.js file with this content:

```js
import InboxSDK from '.';
InboxSDK._setInjectScriptImplementation(() => {
  const script = document.createElement('script');
  script.type = 'text/javascript';
  script.src = browser.runtime.getURL('pageWorld.js');
  document.documentElement.appendChild(script);
});
export default InboxSDK;
```

and a packages/core/firefox.d.ts:

```ts
import InboxSDK from '.';
export default InboxSDK;
```

and we need to make sure that it's guaranteed that the injectScriptImplementation function is only called asynchronously after the InboxSDK import. I think this may already be the case because of an `await import` happening on the path to it in `src/platform-implementation-js/main.ts`, but we can make it explicit by editing `src/platform-implementation-js/dom-driver/gmail/make-xhr-interceptor.ts` to have

```ts
  // Call injectScript asynchronously so InboxSDK alt imports like firefox.js can patch it first
  var pageCommunicatorPromise = Promise.resolve().then(injectScript).then(() => pageCommunicator);
```

(A cleaner solution that didn't need the asynchronous call to injectScript would be for us to not bundle the InboxSDK in our npm release, and then the firefox.js module could import a module that overrides the inject script function before importing and re-exporting the main InboxSDK entrypoint. But it's not so much cleaner that we need to jump to that now.)

DaPotatoMan:
@Macil Taking your suggestion into account, I've removed all `setInjectScriptImplementation` reliance from the code and added a `@inboxsdk/core/firefox` import

So userland usage will now work like this (firefox content script):
```ts
import InboxSDK from '@inboxsdk/core/firefox'
```

<details>
  <summary>How it works</summary>
  
  #### Background
  In the background.js script the `inboxsdk__injectPageWorld` event is sent back to the sender tab (content script) if the browser is firefox
  https://github.com/InboxSDK/InboxSDK/blob/21d765de574da9a0c838997cf466ef40971e888d/packages/core/background.js#L5-L12
  
  #### Content script
  I've added inboxsdk-FIREFOX.ts file which listens for the event and injects pageWorld.js
  https://github.com/InboxSDK/InboxSDK/blob/21d765de574da9a0c838997cf466ef40971e888d/src/inboxsdk-js/inboxsdk-FIREFOX.ts#L1-L21
  
  This method eliminates the need for `setInjectScriptImplementation` usage and further simplifies the implementation.
</details>


<details>
  <summary>Additional notes</summary>
  
  #### Build config
  I've added the firefox entry to the `gulpfile.babel.ts` here, 
  But I'm unsure if I have to add the entry in other places as well.
  https://github.com/InboxSDK/InboxSDK/blob/21d765de574da9a0c838997cf466ef40971e888d/gulpfile.babel.ts#L402-L409
</details>

Please let me know if these changes satisfy your concerns.

src/common/extension-apis.ts
DaPotatoMan: It's improbable that it will cause an error. I don't have a mac so I'm unsure about Safari <14 versions.

wegry: fwiw InboxSDK's .browserslistrc notes that we only support back to Safari 15.4. I'd like to only support Safari 16+, but for now, our main product still supports Safari 15.4.

packages/core/background.js
```diff
@@ -1,9 +1,19 @@
-/* global chrome */
-chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
-  if (message.type === 'inboxsdk__injectPageWorld' && sender.tab) {
-    if (chrome.scripting) {
+/** @type {import('webextension-polyfill').Browser} */
+const browser = globalThis.chrome || globalThis.browser;
+const isFirefox = 'browser' in globalThis;
```
wegry, line 3: `'browser'` is present in Chrome, Edge, and Safari versions we officially support.

gulpfile.babel.ts
```diff
@@ -397,6 +399,14 @@ if (args.remote) {
...
         import: './src/inboxsdk-js/inboxsdk-NONREMOTE',
       },
+      firefox: {
+        library: {
+          export: 'default',
+          name: 'InboxSDK',
+          type: OutputLibraryType.UMD,
+        },
+        import: './src/inboxsdk-js/inboxsdk-FIREFOX',
+      },
```
wegry, line 409:
```suggestion
```

I would prefer that we didn't add a new entry point for Firefox as our package is currently quite large.

Instead, it would be on Firefox extension developers to include both `@inboxsdk/core/firefox.js` and `@inboxsdk/core/inboxsdk.js`.

wegry (changes requested):

gulpfile.babel.ts
Macil: Requiring an application to import something of ours for side effects feels gross.

I do agree that building firefox.js with webpack like this would create a lot of duplicate content in our package as it and our main entry point would contain a lot of the same bundle, and our build time would increase a bit from that.

We can support allowing applications to just import `import InboxSDK from '@inboxsdk/core/firefox';` without adding a new webpack entry point by not building firefox.js with webpack. (And in the future maybe we won't use webpack at all for any parts and deliver everything unbundled, not just firefox.js.)

packages/core/background.js
DaPotatoMan: Changed the code to this:
```ts
/Firefox/.test(navigator.userAgent)
```

wegry (approved):
LGTM.

wegry:
I think what we're waiting for prior to a merge is @Macil will sign off on this as well.

andrejslogins:
If see it right, then https://bugzilla.mozilla.org/show_bug.cgi?id=1736575 looks like has been resolved, so this could be iterated upon in order not to require an additional import, correct?

Acibi:
Is there any update about this PR?

rvaidun:
+1, any updates on what needs to be done to get this merged?

DaPotatoMan:
Added support for both newer and legacy Firefox.
InboxSDK can be loaded with just a single import.
```js
import * as InboxSDK from '@inboxsdk/core/firefox';
```

I'll request a review after I go through all the changes again.

nfacha:
Hi
Any update on this PR?

PR #1170, opened 2024-04-08
chore: stop inlining most libraries in the npm published version

### LINKED ISSUE #981: @inboxsdk/core should not inline various libraries
@inboxsdk/core should use something like webpack's externals. For widely used libraries like `lodash`, using the importing project's library versions allows updates, code sharing, and future tooling improvements (in the vein of #905) to occur down the line.

We have a a number of libraries that are currently being inlined.

```js
[
          '@babel/runtime',
          '@macil/simple-base-converter',
          'asap',
          'auto-html',
          'bignumber.js',
          'bimapcache',
          'buffer',
          'classnames',
          'closest-ng',
          'combokeys-capture',
          'contain-by-screen',
          'ext-corb-workaround',
          'fast-deep-equal',
          'immutability-helper',
          'kefir',
          'kefir-bus',
          'kefir-cast',
          'kefir-stopper',
          'live-set',
          'lodash',
          'map-indexed-xf',
          'matches-selector-ng',
          'order-manager',
          'page-parser-tree',
          'pdelay',
          'querystring-es3',
          'react',
          'react-dom',
          'react-draggable-list',
          'react-smooth-collapse',
          'sha.js',
          'ud',
          'ud-kefir',
          'tag-tree',
          'transducers.js',
      ]
```

# Drawbacks

#785 support. Externals make packaging code for users wanting to avoid bundlers trickier.

wegry:
We now depend on the importing project's dependencies. Common dependencies like `lodash` being shared makes the package much lighter. It also highlights some older libraries we should probably remove.

This is a potentially breaking change for `@inboxsdk/core` consumers. Among other things:

- `@inboxsdk/core` is now published as ESM only.
- The UMD default export has been removed from `@inboxsdk/core`. Use `import * as InboxSDK from '@inboxsdk/core'` instead.

Closes #981.


Macil:
I assume this is intended as a stepping stone toward not bundling the InboxSDK at all.

It seems like this would switch us to publishing using ESM instead of CommonJS. Could this break users who were doing `import InboxSDK from '@inboxsdk/core';` (instead of `import * as InboxSDK from '...';`)? Does that work currently? If it did, we would need to decide whether to document that as a breaking change or to add a default export to make that still work.

wegry:
> I assume this is intended as a stepping stone toward not bundling the InboxSDK at all.

@Macil that's my intention. Ultimately, I think, we'd want a named exports for what are currently "namespaces" in the `InboxSDK` global.

> It seems like this would switch us to publishing using ESM instead of CommonJS. Could this break users who were doing import InboxSDK from '@inboxsdk/core'; (instead of import * as InboxSDK from '...';)?

This actually would break in the reverse in 994571d2ec679e1242fa09dae6b4edbd584dbbaa. `import InboxSDK from '@inboxsdk/core'` (default import) would work. `import * from '@inboxsdk/core` (wildcard import) would stop working. 

> Does that work currently? If it did, we would need to decide whether to document that as a breaking change or to add a default export to make that still work.

Yeah, this is a breaking change. My understanding is that CJS being consumed by a bundler with support for it will work with either a default or wildcard import. Switching to ESM from UMD also removes the `InboxSDK` global provided by webpack's `entry.library.name`.

gulpfile.babel.ts
```diff
@@ -136,6 +139,8 @@ const enum OutputLibraryType {
 
 interface WebpackTaskOptions {
   entry: webpack.Configuration['entry'];
+  externals?: webpack.Configuration['externals'];
+  externalsType?: webpack.Configuration['externalsType'];
```
wegry, line 143:
```suggestion
```

wegry:
This PR may sit until Streak moves off of MV2 SDK support.

PR #1174, opened 2024-04-16
chore(ButtonView): add iconLiga

wegry:
Allow users to specify a default ButtonView.innerText for use with iconClass and icon fonts

### DIFF
--- a/src/inboxsdk.d.ts
+++ b/src/inboxsdk.d.ts
@@ -416,12 +416,12 @@ export { type MoleOptions as MoleDescriptor };
 export interface ButtonDescriptor {
   title?: string;
   /**
-   * The innerText of the button.
-   * If you want to use an icon font with ligatures enabled, you can use the `iconLigaDefault` property.
+   * The innerText of the button when {@link ButtonDescriptor#iconClass} is provided.
+   * If you want to use an icon font with ligatures enabled, you can use the `iconLiga` property.
    *
    * @default '&nbsp;'
    */
-  iconLigaDefault?: string;
+  iconLiga?: string;
   iconUrl?: string;
   iconClass?: string;
   hasDropdown?: boolean;
--- a/src/inboxsdk.d.ts
+++ b/src/inboxsdk.d.ts
@@ -415,6 +415,13 @@ export interface DrawerDescriptor {
 export { type MoleOptions as MoleDescriptor };
 export interface ButtonDescriptor {
   title?: string;
+  /**
+   * The innerText of the button.
+   * If you want to use an icon font with ligatures enabled, you can use the `iconLigaDefault` property.
+   *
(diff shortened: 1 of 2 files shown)


src/platform-implementation-js/dom-driver/gmail/widgets/buttons/button-view.ts
```diff
@@ -41,11 +43,12 @@ export default class ButtonView implements ButtonViewI {
   private _eventStream: Bus<any, any>;
 
   constructor(options: ButtonViewOptions) {
+    const { iconLigaDefault, iconClass } = options;
     this._hasDropdown = false;
     this._isEnabled = options.enabled !== false;
 
-    this._iconClass = options.iconClass;
-    this._iconUrl = options.iconUrl;
```
ruthhalteman, line 48: why is this line removed, should it be there in addition not instead?

wegry: Good catch. See [3f42061](https://github.com/InboxSDK/InboxSDK/commit/3f420617fcc7285be1bdbb93c51de7c6841cacec)

src/inboxsdk.d.ts
```diff
@@ -415,6 +415,13 @@ export interface DrawerDescriptor {
 export { type MoleOptions as MoleDescriptor };
 export interface ButtonDescriptor {
   title?: string;
+  /**
+   * The innerText of the button.
+   * If you want to use an icon font with ligatures enabled, you can use the `iconLigaDefault` property.
+   *
+   * @default '&nbsp;'
+   */
+  iconLigaDefault?: string;
```
ruthhalteman, line 424: what's the reasoning for having "default" be part of this property name? it's not really a default, it's just setting something, so it was a bit confusing to me

wegry: I'm not attached to the name by any means. Would this be more your speed?

```suggestion
  iconLiga?: string;
```

The TSDoc could be clearer that we only set this property if `iconClass` is also passed.

src/inboxsdk.d.ts
```diff
@@ -415,6 +415,13 @@ export interface DrawerDescriptor {
 export { type MoleOptions as MoleDescriptor };
 export interface ButtonDescriptor {
   title?: string;
+  /**
+   * The innerText of the button.
+   * If you want to use an icon font with ligatures enabled, you can use the `iconLigaDefault` property.
```
ruthhalteman, line 420:
```suggestion
   * If you want to use an icon font with ligatures enabled, you can use the `iconLiga` property.
```

wegry: See  5985a5d

ruthhalteman (approved):

PR #1175, opened 2024-04-17
chore(platform-implementation): drop inboxBeta opt

wegry:
- Remove dead `inboxBeta` option that was tied to #887.
- Expose `PiOpt` type and run through usages in `Modal`, `Logger`, and `Global`

### DIFF
--- a/src/inboxsdk-js/inboxsdk.ts
+++ b/src/inboxsdk-js/inboxsdk.ts
@@ -2,6 +2,7 @@ import { PlatformImplementationLoader } from './loading/platform-implementation-
 import { BUILD_VERSION } from '../common/version';
 import { loadScript as _loadScript } from './load-script-proxy';
 import type { LoadScriptOptions } from '../common/load-script';
+import type { PiOpts } from '../platform-implementation-js/platform-implementation';
 
 declare global {
   var __test_origin: string | undefined;
@@ -20,7 +21,7 @@ export function loadScript(
   return _loadScript(url, opts);
 }
 
-export function load(version: any, appId: string, opts: any) {
+export function load(version: number, appId: string, opts?: Partial<PiOpts>) {
   opts = Object.assign(
     {
       // defaults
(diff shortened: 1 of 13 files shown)


src/platform-implementation-js/platform-implementation.ts
```diff
@@ -73,16 +72,19 @@ export class PlatformImplementation extends SafeEventEmitter {
...
-    const members = {
-      driver,
-    };
-    memberMap.set(this, members);
+    this.#driver = driver;
 
     if (process.env.NODE_ENV !== 'production') {
-      (this as any)._members = members;
+      this._members = { driver };
     }
```
Macil, line 88: let's axe this. (Chrome lets you see private variables in the console which addresses this use-case.)

src/platform-implementation-js/platform-implementation.ts
```diff
@@ -37,15 +34,17 @@ import isValidAppId from './lib/is-valid-app-id';
...
+  /**
+   * @deprecated Is this used?
+   */
+  appVersion?: string;
+  suppressAddonTitle?: string | null;
   VERSION: string;
   globalErrorLogging: boolean;
-  eventTracking: boolean;
-  inboxBeta: boolean;
+  eventTracking?: boolean;
```
Macil, line 47: This shouldn't be optional. We take a `Partial<PiOpts>` from the user and construct a `PiOpts` object with defaults filling in for any missing required properties including this one.

src/platform-implementation-js/lib/logger.ts
```diff
@@ -272,7 +273,7 @@ function _extensionLoggerSetup(
 
   _extensionLoaderVersion = loaderVersion;
   _extensionImplVersion = implVersion;
-  _extensionUseEventTracking = opts.eventTracking;
+  _extensionUseEventTracking = opts.eventTracking ?? false;
```
Macil, line 276: Following my previous comment that PiOpts.eventTracking shouldn't be optional, this change should be unnecessary too.

src/inboxsdk-js/inboxsdk.ts
```diff
@@ -33,9 +34,9 @@ export function load(version: any, appId: string, opts: any) {
       VERSION: LOADER_VERSION,
       REQUESTED_API_VERSION: version,
     },
-  );
+  ) satisfies PiOpts;
```
Macil, line 37: This should still be a `Partial<PiOpts>` because we still set some default properties further along at `makePlatformImplementation()`.

Oh, I see that `PlatformImplementationLoader.load()` is typed as taking in a `PiOpts`. We should make sure the type is set to `Partial<PiOpts>` up to it getting into `makePlatformImplementation()`.

Macil (approved):

PR #1176, opened 2024-04-18
tweak selector for last message footer button adding

ruthhalteman:
Change the selector so that it accepts the inline compose or the button version of the bottom element, and then can set the observer up more properly

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
@@ -685,7 +685,7 @@ class GmailThreadView {
 
   addFooterButton(button: ButtonDescriptor) {
     // This element contains either an inline reply or the row of buttons
-    const lastMessageFooterSelector = 'div.nH .aHU div[role="list"] .gA';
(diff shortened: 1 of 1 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
```diff
@@ -686,10 +686,10 @@ class GmailThreadView {
   addFooterButton(button: ButtonDescriptor) {
     // This element contains either an inline reply or the row of buttons
     const lastMessageFooter = this._element.querySelector(
-      '.aHU.hx div[role="list"] .gA',
+      'div.nH .aHU div[role="list"] .gA',
     );
     if (!lastMessageFooter) {
-      throw new SelectorError('.aHU.hx div[role="list"] .gA', {
+      throw new SelectorError('div.nH .aHU div[role="list"] .gA', {
```
wegry, line 692: Sharing the selector between the query might be better here.

```suggestion
    const selector = 'div.nH .aHU div[role="list"] .gA';
    const lastMessageFooter = this._element.querySelector(
       selector,
     );
     if (!lastMessageFooter) {
       throw new SelectorError(selector, {
```

wegry (approved):

PR #1178, opened 2024-04-19
Fix npm build source maps

Macil:
1. It seems like we forgot to enable sourcemaps on our production npm build. (If args.production was on, we wouldn't make source maps, and if it was off, we'd make inline source maps. I've just made it always use external source maps now. I think I might have previously configured some things to use inline source maps in dev because I wasn't sure how to make external source map files work in an MV3 extension, but I've since figured out they work fine if you list them in the manifest's web_accessible_resources, as I've updated the [hello-world example extension](https://github.com/InboxSDK/hello-world) to.) (The "integrated page world" build still uses inline source maps because the workflow it was built for involved copying the inboxsdk.js file alone around. I think that build won't be useful at all once we're fully MV3 so I'm just not touching it.)
2. We weren't using source-map-loader, so the source map comments in some of our bundled dependencies were making it into our output files and would confuse the browser. (Ideally we wouldn't bundle dependencies or anything and wouldn't need to use source-map-loader, but I just want to fix one thing at a time here.)

I've tested with https://github.com/InboxSDK/hello-world that the InboxSDK source maps work now.

### DIFF
--- a/gulpfile.babel.ts
+++ b/gulpfile.babel.ts
@@ -162,6 +162,12 @@ async function webpackTask({
           type: 'asset/source',
         },
         {
+          test: /\.m?js$/,
+          include: /node_modules/,
+          enforce: 'pre',
+          use: ['source-map-loader'],
+        },
+        {
           exclude: /(node_modules|dist|packages\/core)/,
           test: /\.m?[jt]sx?$/,
           use: {
@@ -352,7 +358,8 @@ const sourceMapPlugin = new webpack.SourceMapDevToolPlugin({
 if (args.remote) {
   config = {
     ...remoteCompatConfig,
-    devtool: false,
+    devtool: false, // using manually-configured plugin instead
+    plugins: [sourceMapPlugin],
     entry: {
       [sdkFilename]: {
         library: {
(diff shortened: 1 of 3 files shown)


gulpfile.babel.ts
```diff
@@ -161,6 +161,12 @@ async function webpackTask({
           resourceQuery: /raw/,
           type: 'asset/source',
         },
+        {
+          test: /\.m?js$/,
+          include: /node_modules/,
+          enforce: 'pre',
+          use: ['source-map-loader'],
```
wegry, line 168: I think we should defer on this in favor of #1170 but I won't block on this.

wegry (approved):

gulpfile.babel.ts
```diff
@@ -352,7 +358,8 @@ const sourceMapPlugin = new webpack.SourceMapDevToolPlugin({
 if (args.remote) {
   config = {
     ...remoteCompatConfig,
-    devtool: false,
+    devtool: false, // using manually-configured plugin instead
+    plugins: [sourceMapPlugin],
```
wegry, line 362: This shouldn't affect our sentry source map setup, but roughly an hour after a remote deploy, we'll want to check that source map reporting is working as expected there.

Post MV3 at Streak, our source map setup will be simplified as well.

Macil: I didn't actually make a real change there, I just reordered some lines and added a comment.

wegry: My mistake.

Issue #794, opened 2022-10-13
Question: 


hlkavya0213:
Hi team,

We have a Chrome extension [https://chrome.google.com/webstore/detail/fyle-expense-management/abggpefphmldapcoknbcaadbpdjjmjgk ](https://chrome.google.com/webstore/detail/fyle-expense-management/abggpefphmldapcoknbcaadbpdjjmjgk)  which is used to submit business expenses.
We were able to get the file attachments from the emails earlier using the `getDownloadURL() `method since it's deprecated we can no longer get the attachments without user interaction. 
We figure out a way to do this using the `_attachmentCardImplementation._element` of AttachmentCardClickEvent options. 
Can you please confirm if this is okay to use the `_element` options to get the download URLs?
**Also is there any chance` getDownloadURL() `will be back in the later updates?**

Thanks,
Kavya H L

Macil:
We originally deprecated the `getDownloadURL()` method because in Google Inbox it wasn't possible for us to get the download URL without user interaction and we didn't want to support APIs that wouldn't work in Inbox too, but given that Google Inbox is gone and it looks like it's still possible to get the URL without user interaction in Gmail, there's little reason for us to not bring back `getDownloadURL()`. We can bring it back.

We don't recommend depending on any undocumented underscore-prefixed properties as they weren't intentionally exposed and incompatible changes may happen in them.

hlkavya0213:
Thank you for the reply @Macil, this was really helpful. Is there any specific timeline or ETA to bring back the `getDownloadURL()`? This would really help us further to make any product decisions.

Thanks,
Kavya H L

abhishek1234321:
@Macil If we were to give a PR for this and if it gets approved, would we be able to fast-track and merge it in?

Macil:
Yes, we would be open to taking a PR for this.

PR #1180, opened 2024-04-22
chore: bringing back getDownloadUrl on AttachmentCardView since google inbox is gone

Aniruddha-Shriwant:
`getDownloadURL` was removed from AttachmentCardView here: https://github.com/InboxSDK/InboxSDK/commit/c43493ebc71ce86f303ad14e990dc880ea9a4fda

but this was done for google inbox, which isn't present now, so we discussed this if we could bring it back, discussion link: https://github.com/InboxSDK/InboxSDK/issues/794#issuecomment-1279060304

### DIFF
--- a/src/platform-implementation-js/views/conversations/attachment-card-view.ts
+++ b/src/platform-implementation-js/views/conversations/attachment-card-view.ts
@@ -44,21 +44,7 @@ export default class AttachmentCardView extends (EventEmitter as new () => Typed
     return this.#attachmentCardImplementation.getTitle();
   }
 
-  /**
-   * @deprecated  Please use the same-named method on the AttachmentCardClickEvent object instead.
-   */
   getDownloadURL(): Promise<string | null | undefined> {
-    this.#driver
-      .getLogger()
-      .deprecationWarning(
-        'AttachmentCardView.getDownloadURL',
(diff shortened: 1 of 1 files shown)


wegry (approved):
Thanks for this.

src/platform-implementation-js/views/conversations/attachment-card-view.ts
```diff
@@ -44,21 +44,7 @@ export default class AttachmentCardView extends (EventEmitter as new () => Typed
     return this.#attachmentCardImplementation.getTitle();
   }
 
-  /**
-   * @deprecated  Please use the same-named method on the AttachmentCardClickEvent object instead.
-   */
   getDownloadURL(): Promise<string | null | undefined> {
```
wegry, line 47:
```suggestion
  /**
   * Get the URL for the attachment card's download link as a promise for a string.
   * For FILE attachment cards, the URL will be a short-lived URL that can be 
   * accessed without cookies. For CUSTOM attachment cards, the URL will be the    
   * downloadUrl property of the card's download button if it has one, otherwise 
   * null. Other attachment card types may not have a download URL, and the promise 
   * may resolve to null.
   */
```

wegry:
Released in [@inboxsdk/core@2.1.52](https://github.com/InboxSDK/InboxSDK/releases/tag/v2.1.52)

PR #1181, opened 2024-04-22
chore: build examples like InboxSDK/hello-world

wegry:
Build this repo's bundled examples more similarly to InboxSDK/hello-world, but using esbuild. This allows us to:

- Test our source map setup, minify, and production builds in MV3
- `import * as InboxSDK from '@inboxsdk/core` in our example content scripts (bonus: type hints in JS, and we avoid using the UMD `InboxSDK` global
- write example extensions in TypeScript

Acts as a stepping stone for #1170

### DIFF
--- a/examples/compose-dropdown-button/manifest.json
+++ b/examples/compose-dropdown-button/manifest.json
@@ -5,7 +5,7 @@
   "content_scripts": [
     {
       "matches": ["https://mail.google.com/*"],
-      "js": ["inboxsdk.js", "content.js"],
+      "js": ["dist/content.js"],
       "css": ["content.css"],
       "run_at": "document_end"
     }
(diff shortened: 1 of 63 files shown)


gulpfile.babel.ts
```diff
@@ -29,7 +30,10 @@ const args = stdio.getopt({
     key: 'c',
     description: 'Copy dev build to Streak dev build folder',
   },
-});
+  examples: {
```
wegry, line 33: I debated whether or not to include this flag. Examples take 1s to build, but I figured making them opt in was more explicit.

examples/compose-dropdown-button/manifest.json
```diff
@@ -14,7 +14,7 @@
   "web_accessible_resources": [
     {
       "matches": ["https://mail.google.com/*"],
-      "resources": ["pageWorld.js", "monkey.png"]
+      "resources": ["pageWorld.js", "monkey.png", "*.map"]
```
wegry, line 17: Borrows this trick from https://github.com/InboxSDK/hello-world/commit/83f1168fb5d08f056cb7d7a86c4ec532bf2dacf3#diff-c0eb04b279c83eb3e6113c1ec2beeac82c6526bc38ba5287f1a59d236204ba14R26

PR #1183, opened 2024-04-26
fix(thread view): tweak notice bar spacing with ai suggestions container

dmy7r0:
## Description

Tweak the notice bar bottom spacing when AI suggestions ("Summarize this email" etc.) are enabled in the thread view. 

Example:
| Before the fix      | After the fix |
| ----------- | ----------- |
| image  | image |

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
@@ -211,6 +211,10 @@ class GmailThreadView {
     '2023_11_16': '* > .nH',
   };
 
+  _subjectAISuggestionsContainerSelectors = {
+    '2024_04_26': '.nH > .einvLd',
+  };
+
   addNoticeBar(): SimpleElementView {
     const el = document.createElement('div');
     el.className = idMap('thread_noticeBar');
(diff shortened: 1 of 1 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
```diff
@@ -243,6 +247,30 @@ class GmailThreadView {
...
+      const hasAIButtonsSubjectContainer =
+        !!this._element.querySelector(selector);
+
+      if (!hasAIButtonsSubjectContainer) {
+        continue;
+      }
+
+      this._driver
+        .getLogger()
+        .eventSdkPassive('addNoticeBar AIButtonsSubjectContainer', {
```
jfmaggie, line 264: do you intend to break out the event name here?
```suggestion
        .eventSdkPassive('addNoticeBarAIButtonsSubjectContainer', {
```

jfmaggie (approved):
one minor comment, lgtm

jfmaggie:
oh make sure you double check when reading pane is enabled case: (it should work just to keep this in mind when working with thread related changes :) )

gmail settings
- inbox
- enable reading pane
- right/bottom show the pane

src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
```diff
@@ -211,6 +211,10 @@ class GmailThreadView {
     '2023_11_16': '* > .nH',
   };
 
+  _subjectAISuggestionsContainerSelectors = {
```
wegry, line 214:
```suggestion
  #subjectAISuggestionsContainerSelectors = {
```

src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
dmy7r0: Good question! I was following the code style above (line 246):
Screenshot 2024-04-26 at 2 05 19 PM

I could keep it as is for consistency or swap both but the analytics reporting might break if we use it somewhere.

jfmaggie: gotcha, no need to change, i was just curious if you were following the pattern anywhere.

src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
dmy7r0: Yeah, I see your point however I followed the pattern for consistency so I'd rather refactor this file separately to keep the PR footprint minimal.

PR #1184, opened 2024-04-26
feat: add iconLiga option for right sidebar

wegry:
In `AppSidebar`, allow the same sort of `iconLiga` setup that was added in #1174.

The third icon down in both screenshots uses the new option. The extension I tested with was `sidebar-example`.

Screenshot 2024-04-26 at 14 17 08
Screenshot 2024-04-26 at 14 14 31

### DIFF
--- a/examples/sidebar-example/content.js
+++ b/examples/sidebar-example/content.js
@@ -1,5 +1,18 @@
 import * as InboxSDK from '@inboxsdk/core';
 
+function addStyle(href) {
+  const linkEl = document.createElement('link');
+  linkEl.rel = 'stylesheet';
+  linkEl.href = href;
+
+  if (!document.head) throw new Error('missing head');
+  document.head.appendChild(linkEl);
+}
+
+addStyle(
+  'https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@20..24,400..500,0,0',
+);
+
 InboxSDK.load(2, 'sidebar-example', {
   appName: 'Twitter',
   appIconUrl:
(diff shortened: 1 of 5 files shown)


examples/sidebar-example/content.js
```diff
@@ -1,5 +1,18 @@
 import * as InboxSDK from '@inboxsdk/core';
 
+function addStyle(href) {
```
ruthhalteman, line 3: I kind of wonder if this is making the example extension more confusing with this extra stuff. The example extensions don't have to demonstrate every different configuration.

src/platform-implementation-js/dom-driver/gmail/views/gmail-app-sidebar-view/primary/index.tsx
```diff
@@ -374,9 +378,19 @@ class GmailAppSidebarPrimary {
         buttonContainer = document.createElement('div');
         buttonContainer.className = idMap('sidebar_button_container');
         buttonContainer.setAttribute('data-app-name', appName);
-        buttonContainer.innerHTML = autoHtml`
-          <button class="inboxsdk__button_icon ${iconClass}" type="button" data-tooltip="${appName}">
-            <img class="inboxsdk__button_iconImg" src="${iconUrl}">
+        buttonContainer.innerHTML = `
+          <button class="inboxsdk__button_icon ${escape(
+            iconClass,
+          )}" type="button" data-tooltip="${escape(appName)}">
```
ruthhalteman, line 384: would changing it to escaping these cause any issues for existing usages?

wegry: I don't think so. `auto-html` doesn't support conditionals, but it does use `_.escape` under the covers for values. This is just falling back to that level.

examples/sidebar-example/content.js
wegry: We're a long ways from every configuration, but testing that the current change works doesn't have to be something we ultimately commit. I can back this file out if you'd like.

ruthhalteman: yeah, I was just trying to imagine I was someone new to the SDK looking at the examples with the view that they're the concise "right" way to use the basic version of the features. This stuff adds more than a little bit of extra noise that doesn't have to do with the SDK

PR #1185, opened 2024-04-26
chore: true private in GmailThreadView

wegry:
ViewDrivers typically aren't exposed to the outside API, and so most fields that are marked with the old convention around  prefixes for non-public are straightforward.


src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-view.ts
```diff
@@ -908,99 +906,73 @@ class GmailThreadView {
...
-    this._eventStream.emit({
+    this.#eventStream.emit({
       type: 'internal',
       eventName: 'messageCreated',
       view: messageView,
     });
   }
 
-  _setupWidthManager() {
-    let widthManager = this._widthManager;
```
wegry, line 974: This method doesn't ever seem to be called. With true private used for the method name, `#` prefixed, the TS Server points out that it can be dropped.

This is the only instance of `WidthManager` which can also be dropped.

Issue #1190, opened 2024-05-01
Drop `closest-ng` library


wegry:
This function has a native method equivalent: https://caniuse.com/element-closest. We currently only use it in ~3 spots.

Related to #907 and #1170.

PR #1197, opened 2024-05-17
chore: Drop closest-ng library

ozipi:
- Removed closest imports
- Removed the first param and leave the classname as the only one

Closes https://github.com/InboxSDK/InboxSDK/issues/1190

### DIFF
--- a/src/injected-js/setupCustomViewEventAssassin.ts
+++ b/src/injected-js/setupCustomViewEventAssassin.ts
@@ -42,7 +42,9 @@ function shouldBlockEvent(event: KeyboardEvent): boolean {
 
   // Block all escape key presses inside a custom view, even when an input
   // is focused.
-  if (event.key === 'Escape' && target.closest('.inboxsdk__custom_view')) {
+  const target = event.target;
+  // ...
+  if (event.key === 'Escape' && target instanceof HTMLElement && target.closest('.inboxsdk__custom_view')) {
     return true;
   }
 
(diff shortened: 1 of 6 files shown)


src/injected-js/setupCustomViewEventAssassin.ts
```diff
@@ -43,7 +42,7 @@ function shouldBlockEvent(event: KeyboardEvent): boolean {
 
   // Block all escape key presses inside a custom view, even when an input
   // is focused.
-  if (event.key === 'Escape' && closest(target, '.inboxsdk__custom_view')) {
+  if (event.key === 'Escape' && target.closest('.inboxsdk__custom_view')) {
```
wegry, line 45: Above here, we have a type assertion for `target` to be a  `HTMLElement`. To avoid a potential `TypeError`, we should remove that type assertion, and check in this if block for something like:

```suggestion
  const target = event.target;
  // ...
  if (event.key === 'Escape' && target instanceof HTMLElement && target.closest('.inboxsdk__custom_view')) {
```

wegry (changes requested):
Hey, thanks for this. I think it's quite close, there's just the one comment on guarding against a `null | undefined` `event.target`.

ozipi:
@wegry added your suggestion and reverted because we already had a target declaration on line 38 as `const target = event.target as HTMLElement;`

Please let me know if anything else is needed.

wegry:
> added your suggestion and reverted because we already had a target declaration on line 38 as const target = event.target as HTMLElement;

@ozipi My mistake that that suggestion was unclear. Let's remove the existing unsafe type assertion on line 38. It should be `const target = event.target;`.

We'll still want to add an instanceof check. Line 47 would be something like

```ts
  if (event.key === 'Escape' && target instanceof HTMLElement && target.closest('.inboxsdk__custom_view')) {
```

ozipi:
@wegry done! added another check to the other conditionals on lines 74 and 78

wegry (approved):

wegry:
Thanks for the changes.

ozipi:
Thank you @wegry !!

PR #1199, opened 2024-05-20
dropped event-listener-with-options library

### LINKED ISSUE #1198: Drop `event-listener-with-options` library
The native `EventTarget.prototype.addEventListener` allows options in all browsers we support. We use this library in [one spot](https://github.com/InboxSDK/InboxSDK/blob/d6562cf5c54c7c7382bb0e3db4d74f78d46e98ad/src/platform-implementation-js/lib/fromEventsWithOptions.ts#L6).

AlexVCS:
Closes #1198

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "eslint": "^8.41.0",
     "eslint-plugin-deprecate": "^0.7.0",
     "eslint-plugin-react": "^7.32.2",
-    "event-listener-with-options": "^1.0.3",
     "ext-corb-workaround": "^2.0.0",
     "fast-deep-equal": "^3.1.3",
     "fast-glob": "^3.2.5",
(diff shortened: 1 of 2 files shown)


src/platform-implementation-js/lib/fromEventsWithOptions.ts
```diff
@@ -1,20 +1,14 @@
...
-  Options,
-  addEventListener,
-  removeEventListener,
-} from 'event-listener-with-options';
-
 export default function fromEventsWithOptions(
   target: EventTarget,
   eventName: string,
-  options: Options | boolean,
+  options: boolean,
```
wegry, line 6: This doesn't match the calling code's shape that's passed in.

```suggestion
  options: Parameters<typeof target.addEventListener>[2],
```

wegry (changes requested):
As this is, it doesn't pass a TypeScript `tsc` check, but it would work at runtime. I think fixing up the type is all this needs now.

PR #1208, opened 2024-06-17
app menu: handle async activation

bkorobeinikov:
Gmail could set app menu item to active asynchronously, monitor changes to the menu DOM and deactivate all other menu items that don't have the SDK specific active class name.

Same applies to menu panel. Deactivate all other panels that don't have the SDK specific active class name.


jfmaggie (approved):
lgtm

PR #1209, opened 2024-06-18
Cache extension ID, avoid "Extension context invalidated" errors in logs

Macil:
When an extension is reloaded or removed, certain Chrome APIs in its content scripts switch to throwing "Extension context invalidated" errors when called. The InboxSDK is mostly unaffected since it mostly doesn't use these APIs, except when logging an error it tries to get the extension's ID using a Chrome API. This means that the next time the InboxSDK tries to log an error after the extension is reloaded or removed, the InboxSDK ends up interrupting itself with a new error from trying to look up the extension ID. (This doesn't cause an infinite loop because we catch any errors thrown while logging an error and log them through a simpler code path, but it does add a lot of noise in the console that distracts from any actual errors triggered by the extension itself.)

Screenshot 2024-06-18 at 16 24 58

This PR fixes this by:
1. The getExtensionId() function now uses a try-catch and logs any errors itself instead of letting the error be thrown from it, since for all of the uses of the function it's better to have a `null` instead of being interrupted by an error.
2. We call getExtensionId() once on startup and cache its result, so that way if it's called again later after the extension has been reloaded or removed, it still remembers the extension's ID.

### DIFF
--- a/src/common/get-extension-id.ts
+++ b/src/common/get-extension-id.ts
@@ -1,10 +1,19 @@
 export default function getExtensionId(): string | null {
-  const chrome: any = (global as any).chrome;
-  if (chrome?.runtime?.getURL) {
-    return chrome.runtime.getURL('');
-  }
-  if (chrome?.extension?.getURL) {
-    return chrome.extension.getURL('');
+  try {
+    const chrome: any = (global as any).chrome;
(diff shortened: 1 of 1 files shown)


wegry:
Thanks for this.

PR #1218, opened 2024-08-02
Add ListRouteView.hideSearchPageFilterToolbar() method

Macil:
Adds ListRouteView hideSearchPageFilterToolbar() method that hides this top toolbar on search pages:
Screenshot 2024-08-01 at 4 56 40 PM

Currently does nothing on non-search pages (such as Sent) even though some other pages have a visually similar filter toolbar, as those toolbars are structured differently in the document.


src/platform-implementation-js/dom-driver/gmail/views/gmail-route-view/gmail-route-view.ts
```diff
@@ -267,6 +268,19 @@ class GmailRouteView implements RouteViewDriver {
     return gmailResultsSectionView;
   }
 
+  hideFilterToolbar() {
```
jfmaggie, line 271: still need to push up changes of `hideSearchPageFilterToolbar`?

jfmaggie (approved):
lgtm

PR #1220, opened 2024-08-06
support multiple routes matching active nav

jfmaggie:
this is to support multiple routes can match a nav item's active state. 

in example AppMenu, you can go to `#custom-route-4` manually, it should show nav item 1 being highlighted. if you go to nav item 1, it will default to the first route which is `#custom-route-1`

image

image

(already tested Streak usage of this change)

### DIFF
--- a/examples/app-menu/content.ts
+++ b/examples/app-menu/content.ts
@@ -25,11 +25,12 @@ InboxSDK.load(2, 'app-menu').then(async (sdk) => {
     document.head.appendChild(sheet);
   };
 
-  for (const n of [1, 2, 3]) {
+  for (const n of [1, 2, 3, 4]) {
     sdk.Router.handleCustomRoute(`custom-route-${n}`, (customRouteView) => {
       const el = document.createElement('span');
       el.innerHTML = `This is custom route ${n}`;
       customRouteView.getElement().appendChild(el);
+      console.log('custom-route', n);
     });
   }
 
@@ -51,7 +52,7 @@ InboxSDK.load(2, 'app-menu').then(async (sdk) => {
       routeID: 'custom-route-1',
       isRouteActive: (route) => {
         const routeID = route.getRouteID();
-        return routeID === 'custom-route-1';
+        return routeID === 'custom-route-1' || routeID === 'custom-route-4';
       },
     }),
     customItem2 = sdk.AppMenu.addMenuItem({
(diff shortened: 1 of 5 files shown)


src/platform-implementation-js/views/route-view/route-view.ts
```diff
@@ -22,9 +22,9 @@ class RouteView extends EventEmitter {
   }
 
   /**
-   * @returns a string of the ID of the RouteView. This is the same routeID that you give Router.goto() or Router.createLink(). This will be a value from NativeRouteIDs.
+   * @returns a string or string[] of the ID of the RouteView. This is the same routeID that you give Router.goto() or Router.createLink(). This will be a value from NativeRouteIDs.
    */
-  getRouteID(): string {
+  getRouteID(): string | string[] {
```
Macil, line 27: This change would be an API breaking change (applications can no longer pass the return value of this to Router.goto() or Router.createLink()) but also it's unnecessary: a RouteView represents the URL the user is currently at and they will only be at one route at a time. The changes in this file and route-view-driver.ts should be discarded.

examples/app-menu/content.ts
```diff
@@ -128,7 +129,10 @@ InboxSDK.load(2, 'app-menu').then(async (sdk) => {
 
   panel1.addNavItem({
     name: 'Nav Item 1',
-    onClick: () => sdk.Router.goto('custom-route-3'),
+    routeID: ['custom-route-1', 'custom-route-4'],
```
Macil, line 132: This API is a bit awkward because
- not all of the values are used for the same purposes. All of them are used for checking if the nav item is active, but only the first one is used as the default link.
- You can pass multiple route IDs but not multiple route params (and it looks like the implementation entirely ignores route params during nav item active checking if multiple route IDs are passed).

I think a better API would be to add a `secondaryRoutes?: Array<{routeID: string; routeParams?: Record<string, string|number>}>` property to `NavItemDescriptor` which are only used for active checking, and then make the implementation of `_handleRouteViewChange` in nav-item-view.ts be:

```ts
function _handleRouteViewChange(
  navItemViewDriver: GmailNavItemView,
  [navItemDescriptor, routeViewDriver]: [NavItemDescriptor, RouteViewDriver]
) {
  let isActive = false;
  if (navItemDescriptor) {
    const routesToCheck = [
      {
        routeID: navItemDescriptor.routeID,
        routeParams: navItemDescriptor.routeParams,
      },
      ...(navItemDescriptor.secondaryRoutes || []),
    ];
    isActive = routesToCheck.some(
      (route) =>
        routeViewDriver.getRouteID() === route.routeID &&
        isEqual(routeViewDriver.getParams(), route.routeParams || {})
    );
  }
  navItemViewDriver.setActive(isActive);
}
```

Macil (changes requested):

Macil (approved):

PR #1228, opened 2024-11-07
quick fix

jfmaggie:
tested current side panel for a lot of users who don't have the gmail side panel changes yet still works:

image

gmail is rolling out changes to show side panel like this, tested on gmail that has the change works too:
image


Macil:
Does this css work with the old Gmail formatting too?

src/platform-implementation-js/style/gmail.css
```diff
@@ -1806,7 +1806,7 @@ body :local(.sidebar_iconArea) .inboxsdk__button_icon {
   .sidebar_button_container_active
   .inboxsdk__button_selectedIndicator {
   height: 40px;
-  width: 4px;
+  width: 3px;
```
jfmaggie, line 1809: the position of the indicator also changed to be on left side instead of right side

PR #1229, opened 2024-11-13
Tweak positioning to work with both old and new popover style

ruthhalteman:
Gmail has new popover styles for editing links. Fix the positioning of the SDK popovers.

A tad suspicious why I had the top position defined in the first place but maybe there was something going on with that container that they changed to support the new content state (there are still states that use the old style popover)

Before:
image

After:
image

### DIFF
--- a/src/platform-implementation-js/style/gmail.css
+++ b/src/platform-implementation-js/style/gmail.css
@@ -2800,7 +2800,6 @@ table[role='presentation'].inboxsdk__thread_view_with_custom_view > tr {
 .inboxsdk__linkPopOver_section_container {
   position: absolute;
   width: 100%;
-  top: 35px;
   left: 0;
 }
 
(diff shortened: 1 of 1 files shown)


jfmaggie:
can you double check the old style as well?

ruthhalteman:
> can you double check the old style as well?

whoop I missed one of the styles I had tested with but yes I've tested with the old, new, and old in the new styles haha

ruthhalteman:
@jfmaggie 

Full old style (my personal account still has it):
image

"old" style in the new UI flow:
image

new:
image

jfmaggie (approved):
lgtm

jfmaggie:
nit: i think we are getting the parent width as 100%, right? and on the new style, it looks just a little bit wider, not a huge deal, we can tweak to get better styling of course

ruthhalteman:
> nit: i think we are getting the parent width as 100%, right? and on the new style, it looks just a little bit wider, not a huge deal, we can tweak to get better styling of course

I think it's because of the shadow, but yeah. The SDK user has full easy control of the popover styles so it shouldn't be too hard to deal with if desired. I just left it because it obviously doesn't match the new styling anyway. Not sure if we'd come back and change more to match or not

Issue #809, opened 2022-11-27
Bug: When open the gmail with Gmail calendar open, then open the the sidePanel created by inboxSDK.Global.addSidebarContentPanel or threadView.addSidebarContentPanel may got empty page.


fantaomeng:
**When open the gmail with Gmail calendar open, then open the the sidePanel created by inboxSDK.Global.addSidebarContentPanel or threadView.addSidebarContentPanel may got empty page.**

**To Reproduce**

1. Open the gmail with Gmail calendar open.
2. Click the SidePanel created by inboxSDK.Global.addSidebarContentPanel or threadView.addSidebarContentPanel may got empty page sometimes.

**Expected behavior**
Expect to see the target SidePanel, but we got empty page.

**Screenshots**
<img width="472" alt="image" src="https://user-images.githubusercontent.com/88301653/204145605-77ee7f80-c505-496c-8e35-822d6c8fd9b6.png">
<img width="492" alt="image" src="https://user-images.githubusercontent.com/88301653/204145642-77564759-36a3-43e1-a3ce-1ed9344f3587.png">

**We can't reproduce this issue every time**

1TheMuffinMan:
I'm able to reproduce this every time. 
Reproduction steps:

1. Considering you add the side bar after inboxsdk loads.
2. Open the sidebar such as the calendar and leave it open
3. Refresh the page
4. Observe that the problem is reproducible

The hack I've been using is `setTimeout` with a 1500ms delay. I don't think it's perfect. Still seems to randomly not work, but feels 95%.

The way I seemed to have solved it is by using `setTimeout` for 1500ms be

vishnug-bridge:
Any solution for this issue ?

vishnug-bridge:
@1TheMuffinMan can you give me the solution?

zakstal:
@aleemstreak 

I took a look at the issue and followed @1TheMuffinMans steps on the sidebar contact example (`examples/sidebar-contact-example`) and was able to reproduce.

The issue appears to be a race condition. The [gmail-element-getter](https://github.com/InboxSDK/InboxSDK/blob/0de424786e0c05834f753fedaa89e852ce5df139/src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts#L45) is querySelecting the sidebar container using the class `.brC-brG` (sidebar selector) . 

[...]

PR #1232, opened 2024-11-20
Add more specific class to avoid race condition when loading the inbox sdk

zakstal:
# Change log

This is a fix for the bug reported on issue [#809](https://github.com/InboxSDK/InboxSDK/issues/809)

The issue appears to be a race condition. The [gmail-element-getter](https://github.com/InboxSDK/InboxSDK/blob/0de424786e0c05834f753fedaa89e852ce5df139/src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts#L45) is querySelecting the sidebar container using the class `.brC-brG` (sidebar selector) . 

However, if the InboxSDK loads before the gmail calendar, the sidebar selector will select an element with the full class list of `.brC-brG .brC-aW-aV` (first element), which will be removed and replaced by a very similar element with a full class list of `.brC-brG .brC-aMv-auO` (second element).

So, the InboxSDK attaches to the first element which, the calendar loads, and the first element is removed from the DOM and the side panel is forever doomed. 

This does not appear to be happening on the main Streak product because it just loads slower then the calendar.

___Before fix class list change___
before-wrong-class-calload

___After fix class list change___

afterfix-right-class-calload

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/gmail-element-getter.ts
@@ -42,7 +42,7 @@ const GmailElementGetter = {
   },
 
   getCompanionSidebarContentContainerElement(): HTMLElement | null {
-    return document.querySelector('.brC-brG');
+    return document.querySelector('.brC-brG.brC-aMv-auO');
   },
 
   // <div class="brC-aT5-aOt-Jw" role="complementary" aria-label="Side panel">


1TheMuffinMan:
I briefly looked over your explanation and change. You mentioned that the Streak product doesn't face this issue because it loads too slow, so will this change now break their implementation?

zakstal:
@1TheMuffinMan I can't be sure that it will not break their product without actually testing it, but here are some observations.
(using some language from the initial change log like `first element` and `second element`)
- there are never any other elements on the page other the `first element` and the `second element`  with the class `.brC-brG` as far as I can tell. I put an interval in the DOM as soon as the SDK loads using `document.querySelectorAll('.brC-brG')` and I only ever see one of those elements at a time. 
- All the native tabs on first load have the `first element` class list `.brC-brG .brC-aW-aV` on load.
- when running the `examples/sidebar-contact-example` and using your repro steps, stepping through the DOM I can see the InboxSDK attaching to the `first element` then it being replaced by the `second element`. 
- When running the `Streak` product and using the repro steps, I can see the gmail cal loading first every time, then the streak sidebar elements getting attached to the `second element`

Assuming they are using the SDK (which they are) and given that there are no other elements with the class `.brC-brG` and given that this pr fixes `examples/sidebar-contact-example` **I think there is a high probability this won't break the `Streak` product**.

However, classes change, and I don't know how often the additional class will be updated by gmail. So thats the risk.

zakstal:
@Macil Hello, can this be merged? thanks!

Issue #1153, opened 2024-03-20
Bug: Cannot update/reset values via `replaceDate` or `replaceDraftLabel`


fjaeger:
When calling `replaceDraftLabel` or `replaceDate` multiple times with different text values or when trying to reset to the original value of a thread row, the behavior is not as I would expect it. 

* Resetting is not doing anything!
* Updating does not replace the value, but adds another `<span>`, so that we end up with duplicate values.

**Expected behavior**

When passing `null` as the parameter to `replaceDraftLabel` or `replaceDate`, I would expect the corresponding values for the thread row to be reset.
When calling `replaceDraftLabel` or `replaceDate` repeatedly, I would expect the corresponding values for the thread row to be replaced with the latest `text` value.

**Version**

I am using the latest version of InboxSDK, installed/bundled via npm.

wegry:
@fjaeger thanks for the report. Can you create a minimum reproducible example?

fjaeger:
@wegry Sure! Please find the demo project here: https://github.com/Mailbutler/inboxsdk-row-bug-demo

wegry:
For posterity's sake, with the following diff against 9dd4175d and the much older cc1f4f0c4da1314f9d5813a0a5b9193217612680, it looks like this issue wasn't introduced by recent changes to gmail-thread-row-view in #1148 or #1150.

```diff
diff --git a/examples/thread-rows/content.js b/examples/thread-rows/content.js
index 9123776b..0d0a3bca 100644
--- a/examples/thread-rows/content.js
+++ b/examples/thread-rows/content.js
@@ -23,188 +23,9 @@ InboxSDK.load(2, 'thread-rows').then(function (inboxSDK) {
   });
 
   inboxSDK.Lists.registerThreadRowViewHandler(function (threadRowView) {
-    var threadId = threadRowView.getThreadID();
-    //console.log('threadRowView', threadId, threadRowView.getThreadIDIfStable(), threadRowView.getVisibleDraftCount(), threadRowView.getVisibleMessageCount(), threadRowView.getSubject());
-    console.log('got threadRowView', new Date().toString());
-    if (threadRowView.getVisibleMessageCount() == 0) {
-      threadRowView.addLabel(
-        Kefir.fromPromise(threadRowView.getDraftID()).map((id) => ({
-          title: '' + id,
-        })),
-      );
-    }
+    threadRowView.replaceDate({ text: 'First' });
 
-    threadRowView.addImage(
[...]

PR #1233, opened 2024-11-20
Update replaceDate function to operate on a single element for updates.

zakstal:
# Changelog

This is a potential fix for issue [#1153](https://github.com/InboxSDK/InboxSDK/issues/1153).
My comments and reasoning are [here](https://github.com/InboxSDK/InboxSDK/issues/1153#issuecomment-2489519863)

## Changes
- update retrieving of the dateMod from `this._modifications.replacedDate.unclaimed` to also check `this._modifications.replacedDate.claimed` and use the element in `claimed` if it exists.  This ensures that replaceDate will replace the original date and any subsequent calls will  update the replacement.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.ts
@@ -1093,7 +1093,9 @@ class GmailThreadRowView {
           }
         } else {
           if (!dateMod) {
-            dateMod = this._modifications.replacedDate.unclaimed.shift();
+            dateMod = this._modifications.replacedDate.unclaimed.length 
+              ? this._modifications.replacedDate.unclaimed.shift()
+              : this._modifications.replacedDate.claimed.shift();
 
             if (!dateMod) {
               dateMod = {


Macil:
This change doesn't cause `replaceDate(null)` to undo a previous replaceDate call.

zakstal:
Hi @Macil. So the desired behavior for `replaceDate(null)` is to undo the a previous replaceDate call?

Macil:
Yeah, as mentioned in https://github.com/InboxSDK/InboxSDK/issues/1153#issuecomment-2024107621 the goal is to have calls to both `replaceDate` and `replaceDraftLabel` overwrite any previous calls, including a call with a null argument removing the previous replacement. I think the strategy your PR does of reusing the existing modification is workable but misses the case where the new value is null.

zakstal:
Ok thanks, I'll update the PR. If you have thoughts on a better approach I'd love to hear it.

zakstal:
@Macil Ok updated.

### Added
- `replaceDate(null)` now reset to the original date element
-  repeated calls to `replaceDraftLabel(<text>)` will replace the draft label element text on every call
- `replaceDraftLabel(null)` will reset the draft label element to the original "Draft" text.

If you need code to test against I can include it in the pr or on a separate branch for you to test against. 

I also noticed this behavior when replacing the draft label.

draft-label-fix

The number next to the draft label is originally on the right of the "Draft" text.
When the `replaceDraftLabel(<text>)` is applied the number moves to the left side of the element. 
I assume thats not the desire behavior?

Thanks again for reviewing!

zakstal:
@Macil ping!

PR #1238, opened 2025-01-01
add MessageView.getElement

davidgohberg:
@Macil same motivation and reasoning as https://github.com/InboxSDK/InboxSDK/pull/1165:

Gmail has been relatively stable, Google Inbox was discontinued, it's extremely useful to prototype features using direct access to elements outside of the InboxSDK, and some extensions want to do things that the InboxSDK doesn't want to commit to maintaining, so we're much more willing to expose Gmail elements now where it's useful to.


ataliare:
Very helpful! would like to use it, it can help me also. Thanks!

ataliare:
@Macil Hi! When do u plan to release a new version? I really need this :)

PR #1239, opened 2025-01-09
additional getters in linkPopOver

ruthhalteman:
Two functions to help users access contents of gmail's link popover easier. This is helpful with gmail's new UI where the link editing happens in the popover rather than a separate modal, so there are more states and more activity in the popover now.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
@@ -19,6 +19,17 @@ class LinkPopOver extends SafeEventEmitter {
     return this._linkEl;
   }
 
+  getPopOverContainerElement() {
+    return this._popOverEl;
+  }
+
+  getUrlInputElement() {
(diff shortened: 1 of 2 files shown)


Macil (approved):

src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
```diff
@@ -19,6 +19,17 @@ class LinkPopOver extends SafeEventEmitter {
...
 
+  getPopOverContainerElement() {
+    return this._popOverEl;
+  }
+
+  getUrlInputElement() {
+    // this is for the new gmail UI
+    return this._popOverEl.querySelector<HTMLInputElement>(
+      '.qdOxv-K0-wGMbrd[aria-label="Link"]',
+    );
```
jfmaggie, line 30: we should add logging for new gmail UI so we can track new UI ramp ups.

here is [an example](https://github.com/InboxSDK/InboxSDK/pull/1228/files#diff-80ad5f03e67076a465796a98db96198fc999fda4076b8e34b962edb8da673478R711)

PR #1240, opened 2025-01-10
Potentially resolve #1151: add parseOnActivateResult tests

### LINKED ISSUE #1151: Bug: adding `hasMore: true`  for `CustomListDescriptor` does not enable the pagination button in list view
Whenever we have threads exceeding 50, we set `hasMore: true` in the custom list descriptor for the handleCustomListRoute. In my understanding, the pagination should be enabled for this case so that the user would be able to navigate through the rest of the messages. But for this scenario the pagination remains disabled (image below for reference)

Screenshot 2024-03-07 at 7 19 34 PM

Is there something else that should be passed to enable this? We wouldn't be able to calculate the `total` property and opted for the `hasMore` approach based in the [documentation for CustomListDescriptors](https://inboxsdk.github.io/inboxsdk-docs/router/#customlistdescriptor). If the `hasMore` is not responsible for the pagination enablement, what's the usecase for the `hasMore` property? 

Here is a sample snippet of the scenario above
```javascript
inboxSdk.Router.handleCustomListRoute('list-route', (_, _) => {
  return new Promise((resolve, reject) => {
	const threadIds = ...
	resolve({
	  threads: threadIds,
	  hasMore: true
	})
  })
})
```

current version: 2.1.21
Version 119.0.6045.159 (Official Build) (arm64)
macOS: Sonoma 14.1.1

trogers1:
I'm not a user of InboxSDK, but I was perusing the open issues and saw #1151 and it smelled to me like a simple business logic issue. So I added some tests where I was suspicious and I _think_ I found and fixed a couple of little bugs including one that seems like it could be responsible for #1151. Specifically, it looks like it could be a language-translation issue because the logic was doing `null`-checks on variables who's type did not include `null`, but _did_ include `undefined`, which seemed suspicious.

However, someone should probably properly reproduce the issue and see if this resolves it who is an actual user of the software 😅

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/gmail-driver/show-custom-thread-list.test.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/gmail-driver/show-custom-thread-list.test.ts
@@ -404,8 +404,10 @@ test('missing threads in response', async () => {
 
 describe('parseOnActivateResult', () => {
   const fakeLogger = {
+    /* eslint-disable  @typescript-eslint/no-empty-function */
     deprecationWarning: (..._) => {},
     error: (..._) => {},
+    /* eslint-enable @typescript-eslint/no-empty-function */
   } as Logger;
   test("should default to one page since arrays can't be paginated when provided with a results array", () => {
     const result = [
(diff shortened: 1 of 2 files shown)


Macil:
Thanks for taking a look at this, but this doesn't look like the issue. The `total` variable in that case is meant to be the whole threads list length, and the `==`/`!=` checks against null are fine since double-equals considers null and undefined to be the same.

PR #1241, opened 2025-01-21
Add getter for link popover text input

ruthhalteman:
Actually this one would be helpful too. Same reasoning/thing as https://github.com/InboxSDK/InboxSDK/pull/1239 

Also a little tweak to the width in the new gmail UI case.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
@@ -114,6 +114,11 @@ export default function setupLinkPopOvers(
     ({ el, removalStream }: ItemWithLifetime<TagTreeNode<HTMLElement>>) => {
       const popOverEl = el.getValue();
 
+      gmailComposeView
+        .getGmailDriver()
+        .getLogger()
+        .eventSite('link_popOver', popOverEl);
+
       let existingLinkPopOver: LinkPopOver | null = null;
 
       removalStream.onValue(() => {
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
@@ -30,6 +30,13 @@ class LinkPopOver extends SafeEventEmitter {
     );
   }
 
+  getTextInputElement() {
+    // this is for the new gmail UI
+    return this._popOverEl.querySelector<HTMLInputElement>(
+      '.qdOxv-K0-wGMbrd[aria-label="Anchor Text"]',
+    );
+  }
+
   addSection() {
     let containerEl = this._popOverEl.querySelector<HTMLElement>(
       '.inboxsdk__linkPopOver_section_container',
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
@@ -14,10 +14,10 @@ class LinkPopOver extends SafeEventEmitter {
     this._linkEl = linkEl;
     this._popOverEl = popOverEl;
 
-    this.detectNewGmailUI(this._popOverEl);
+    this.labelNewGmailUI(this._popOverEl);
   }
 
-  detectNewGmailUI(popOverEl: HTMLElement) {
+  labelNewGmailUI(popOverEl: HTMLElement) {
     /* 2025-1-21 version with new gmail style popover */
     if (
       popOverEl.classList.contains('VRezbc') &&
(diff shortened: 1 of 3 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
```diff
@@ -30,6 +30,13 @@ class LinkPopOver extends SafeEventEmitter {
     );
   }
 
+  getTextInputElement() {
+    // this is for the new gmail UI
+    return this._popOverEl.querySelector<HTMLInputElement>(
+      '.qdOxv-K0-wGMbrd[aria-label="Anchor Text"]',
```
ruthhalteman, line 36: would `aria-label` be localized? I was initially assuming no but the formatting here makes me think it might be?

src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
```diff
@@ -30,6 +30,13 @@ class LinkPopOver extends SafeEventEmitter {
     );
   }
 
+  getTextInputElement() {
+    // this is for the new gmail UI
```
jfmaggie, line 34: oh sorry i added the comment in wrong PR: https://github.com/InboxSDK/InboxSDK/pull/1239/files#r1924283251, meant to be here

src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
jfmaggie: yes. you can test it by switching to a diff language. i tried Chinese and here is the dom element

```
<input jsname="YPqjbf" type="text" value="" id="c0" class="qdOxv-K0-wGMbrd" jsaction="input:YPqjbf;focus:AHmuwe;blur:O22p3e" aria-label="定位文字" placeholder="文本" autocomplete="off" style="">
```

ruthhalteman: the `jsname` and `data-idom-container-class` etc are the same for both inputs which is annoying

ruthhalteman: okay @jfmaggie changed the approach, do you think this is safer? it's more tied to what the UX is

src/platform-implementation-js/style/gmail.css
```diff
@@ -2802,6 +2802,12 @@ table[role='presentation'].inboxsdk__thread_view_with_custom_view > tr {
   width: 100%;
   left: 0;
   margin-top: 8px;
+  box-sizing: border-box;
+}
+
+/* version with new gmail style popover: match width better */
+.VRezbc.phnWcc .inboxsdk__linkPopOver_section_container {
```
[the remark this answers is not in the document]
jfmaggie: https://github.com/InboxSDK/InboxSDK/pull/1241/commits/78c72b19011e8f50186ad6db44fa0cbd203bf1c5

src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
```diff
@@ -13,6 +13,18 @@ class LinkPopOver extends SafeEventEmitter {
     super();
     this._linkEl = linkEl;
     this._popOverEl = popOverEl;
+
+    this.detectNewGmailUI(this._popOverEl);
```
ruthhalteman, line 17: should the function name communicate that it's making change? detect to me reads that it'd return a boolean or something

src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
jfmaggie: yeah i checked both inputs parents are almost the same except for ids, so this should be better than using the obviously different `aria-label` when language settings are different.

src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
jfmaggie: with all current changes, we have site logging and new class, should be able to find out who has the new UI. resolving this comment

src/platform-implementation-js/dom-driver/gmail/views/gmail-compose-view/setupLinkPopovers.ts
jfmaggie: ah yeah, makes sense. feel free to update to what can reflect better 👍

ruthhalteman: `labelNewGmailUI`?

src/platform-implementation-js/style/gmail.css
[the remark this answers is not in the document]
jfmaggie: ruth and i chatted on this, `.VRezbc.phnWcc` only appears on cases of this new gmail style UI. so not reliable to detect if this is new gmail UI. so revert back to just use css selectors directly

jfmaggie (approved):
lgtm

PR #1256, opened 2025-05-08
Fix pagination for custom lists

Roensby:
https://github.com/InboxSDK/InboxSDK/issues/1151


Roensby:
this seems to work, but the offset returned in `sdk.Router.handleCustomListRoute()` is not correct for pages other than the first

Roensby:
Oh, I stand corrected, the `offset` is a page offset, not a thread offset. So this should work. Not tested with the internal value `'MANY'` which seems to be set whenever `hasMore` is used.

lhford:
also with pagination -- it seems that while offset is the page offset, there is no reliable way to query how many elements there are per page. While 50 is the gmail default, it is a configurable setting in gmail that inboxsdk currently doesn't seem to read in setting `limit`

PR #1257, opened 2025-05-09
fix gmail mole

bkorobeinikov:
when google chat and meet are disabled, the moleParent element has no children initially, unlike when enabled and there is a hidden mole like stub created by gmail


hwfwalton (approved):
lgtm 👍

Issue #1258, opened 2025-05-16
Bug: Message Compose Mole view is not positioned correctly, hovers the side bar


BRS-Robert:
**Describe the bug**

Message Compose Mole view is not positioned correctly.
The screenshots were made by using Streak extension but the same behaviour happens in case of our custom extension as well.

**Reproduction steps**

1. Install Streak 
2. Open side bar by clicking to the Streak icon
3. Click Compose button to open Compose  MoleView

**Expected behavior**

The MoleViews should be located correctly, relative to the side bar.

**Screenshots**

<img width="584" alt="Image" src="https://github.com/user-attachments/assets/e6ee92d3-c9fe-4f72-b2d2-bb816da0fd0d" />

<img width="549" alt="Image" src="https://github.com/user-attachments/assets/f2e34684-9848-4e60-873f-c9767abfc71c" />

**Add any other context about the problem here if anything**

Our custom built MoleView behaves the same like the Message Compose.
Most probably the issue relates to [these lines of code.](https://github.com/InboxSDK/InboxSDK/blob/e78a04d73908410d5d88d4bb99dd434f0ade2fe7/src/platform-implementation-js/style/gmail.css#L501-L504)

PR #1259, opened 2025-05-19
remove mole position overrides

jfmaggie:
looks like we no longer need the sidebar width override now.

close https://github.com/InboxSDK/InboxSDK/issues/1258

before fix:
https://www.loom.com/share/180105b1fd0846e4b2342989b7fbe6a3

after fix:
https://www.loom.com/share/80aac8d3ede949e9aae4cba66b2a6396


bkorobeinikov (approved):
lgtm

PR #1270, opened 2025-07-23
ThreadRowView Button Fixes

tybro0103:
ThreadRowView Button Fixes:
- use `aria-label` instead of `title` on the DOM element so that only the GMail tooltip is shown, and not the native browser tooltip/hint
- updates to `ThreadRowButtonDescriptor.title` will now update the DOM
- allow tooltips to show when button is next to the ⭐ (this happens when "Hover actions" setting is disabled)

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.ts
@@ -84,41 +84,23 @@ const cachedModificationsByRow: WeakMap<HTMLElement, Mods> = defonce(
   () => new WeakMap(),
 );
 
+const BUTTON_CLASS = 'inboxsdk__thread_row_button' as const;
+
 function focusAndNoPropagation(this: HTMLElement, event: Event) {
   this.focus();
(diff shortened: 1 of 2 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-thread-row-view.ts
```diff
@@ -703,29 +694,20 @@ class GmailThreadRowView {
...
             this._expandColumn('col.y5', 26 * starGroup.children.length);
 
-            // Don't let the whole column count as the star for click and mouse over purposes.
-            // Click events that aren't directly on the star should be stopped.
-            // Mouseover events that aren't directly on the star should be stopped and
-            // re-emitted from the thread row, so the thread row still has the mouseover
-            // appearance.
-            // Click events that are on one of our buttons should be stopped. Click events
-            // that aren't on the star button or our buttons should be re-emitted from the
-            // thread row so it counts as clicking on the thread.
```
tybro0103, line 726: Moved still-relevant comments to def of `onClickStarGroup`

jfmaggie (approved):
lgtm

PR #1271, opened 2025-07-30
Fix moles covering our SDK content panels

ruthhalteman:
[Box Link](https://streak.com/a/boxes/agxzfm1haWxmb29nYWVyNAsSDE9yZ2FuaXphdGlvbiIRb2lzbWFpbEBnbWFpbC5jb20MCxIEQ2FzZRiAgMWR-v2TCAw)

The box was filed for task moles but it's really just all moles when our SDK content panels are in use. Also note that some of gmail's own side panels don't do this right (the gemini one at least).

I used AI to help with the Kefir stuff a bit so if there's things that aren't best practices here please lmk. 

Note that gmail doesn't really adjust when the right side narrow toolbar is collapsed so we're matching that behaviour. I think it might just be so that that arrow button to un-collapse is still properly usable?

### DIFF
--- a/src/platform-implementation-js/views/content-panel-view.ts
+++ b/src/platform-implementation-js/views/content-panel-view.ts
@@ -2,6 +2,11 @@ import EventEmitter from '../lib/safe-event-emitter';
 import get from '../../common/get-or-fail';
 import type ContentPanelViewDriver from '../driver-common/sidebar/ContentPanelViewDriver';
 import type TypedEventEmitter from 'typed-emitter';
+import * as Kefir from 'kefir';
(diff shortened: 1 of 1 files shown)


jfmaggie:
unfortunately i don't think this would be sufficient to handle the mole positioning entirely, see the testing: https://www.loom.com/share/dcc748e40cc64118845d2cb79aec5bfa

when you switch to the gmail sidebar item like calendar, the mole covers the sidebar content panel again (we prematurely reset `.dw .jAmAWb` width back to 66px). so i think the root cause is somewhere near how we identify the sidebar content panel and if there is any of our sidebar open or gmail sidebar open. without our sidebar, the mole positions properly.

PR #1294, opened 2026-04-30
fix(xsrf): isRelevantTo should be true if xsrf has changed

GeoffreyParrier:
Hello again, we had a bug today due to the xsrf token being invalid.

In case the page is not refreshed for a long period of time, the xsrf token can be invalid and it should be updated.
Currently, has the token is already set, `isRelevantTo` return false.

This fix return true in case the tokens are not equal.

Not sure about the token lifespan but it seems to be multiple days, the issues happened on a 16 days old token, so it is quite difficult to reproduce unfortunately.


Macil:
Thanks for reporting this and looking into it! We've included a similar fix based on yours in the newest release, 2.2.12.

PR #1302, opened 2026-07-07
fix(compose): tolerate Gmail title-bar Ht-token DOM variant

pnavarrc:
## Issue

`setTitleBarColor` / `setTitleBarText` locate the compose title-bar `<table>` via `.nH.Hy.aXJ table.cf.Ht`. Gmail is A/B-testing a DOM variant that renames the generated `Ht` class token to a hashed `Ht-<hash>` token (e.g. `Ht-ql-thxFLe-aOr`). CSS class matching is on whole tokens, so `table.cf.Ht` stops matching, `querySelectorOrFail` throws, and the caller's compose setup unwinds.

## Fix

Two changes to `gmail-compose-view.ts`:

1. **Support both DOMs** — try the classic selector first, then a token-agnostic structural fallback, using the existing inline `querySelector(a) || querySelector(b)` idiom already used elsewhere in the driver (e.g. `gmail-thread-row-view.ts`). `td.Hm` / `div.Hp` survive in both DOMs. One code path, no branching on which variant is live (an A/B rollout can serve either DOM to the same user).

2. **Degrade gracefully instead of throwing** — coloring/naming the title bar is cosmetic, so when the element can't be found, log via `driver.getLogger().error(...)` and return a no-op cleanup rather than throwing and unwinding the caller's compose setup. This matches the existing "log and stub" precedent used for the compose message-id field. Background lookups are likewise made defensive (skip missing elements). API-misuse cases (inline reply, already-registered) still throw.

## Rollout

Pre-emptive: this ships first, then the Streak extension consumes it via a version bump. Streak also keeps a thin defense-in-depth guard so a cosmetic title-bar call can't strand compose setup even if this logs.


jfmaggie (approved):
lgtm

PR #1308, opened 2026-08-06
feat: add a selector registry with overrides

pnavarrc:
Adds a selector registry to the Gmail driver: named keys that map to an **ordered list** of CSS candidates, plus an optional override map (`PiOpts.selectorOverrides`) that can *prepend* candidates to any key.

Behavior is unchanged when no overrides are passed.

## Why

Gmail changes its DOM without warning and breaks our selectors. Today a one-character selector fix ships through an `@inboxsdk/core` release, then an extension build, then a multi-day Chrome Web Store review — days of breakage for what is often a single string.

A declarative selector map can be supplied as *remote configuration*, which MV3 permits. This PR is the SDK half of that: the registry and its merge, so a later change can feed it a config. It ships nothing user-facing on its own.

The SDK never fetches anything — overrides arrive as a plain value from the extension, which keeps the SDK a pure function of its inputs.

## How the merge behaves

`mergeSelectorOverrides` iterates the bundled defaults, prepending the valid selectors to its own list. Candidates that fail a CSS parse in the current browser are dropped and reported through `onInvalidSelector`.

```ts
mergeSelectorOverrides(
  { 
    'composeView.someKey': ['.foo'] 
  }, 
  { 
    'composeView.someKey': ['.bar', 'INVALID_SELECTOR'],
    'INVALID_KEY': ['.bar']
  }
);
// { 'composeView.someKey': ['.bar', '.foo'] }
```

The `SelectorRegistry` constructor merges the bundled registry with the overrides, defaulting to the bundled registry on error.

## Where it lives

`GmailDriver` owns the registry, and creates an instance in the constructor, passing the corresponding options (`selectorOverrides`, `onInvalidSelector`). Candidates are resolved per instance, so two InboxSDK instances don't share the registry.

Call sites go through the driver to query selectors

```tsx
this.#driver.selectors.querySelectorByKey(el, 'composeView.titleBarTable');
```

## The two keys

A vertical slice to test the concept — `composeView.titleBarTable` and `composeView.titleBarTd`, lifted verbatim from the hand-rolled `||` chains already in `setTitleBarText` and `setTitleBarColor`:

```ts
// before
const titleBarTable =
  this.#element.querySelector('.nH.Hy.aXJ table.cf.Ht') ||
  this.#element.querySelector('.nH.Hy.aXJ table.cf');

// after
const titleBarTable = querySelectorByKey(this.#element, 'composeView.titleBarTable');
```

Both call sites degrade rather than throw, so both use `querySelectorByKey`. The throwing variant `querySelectorByKeyOrFail` exists for sites that want today's `querySelector` semantics.

## Notes

- The registry is immutable after construction, there is no reload. A config change takes effect on the next SDK load.
- We want to ship this to test if the Chrome Store accepts the selector registry as configuration, once that has passed a few reviews (🤞) we can start adding more selector keys and values over time.

### DIFF
--- a/src/platform-implementation-js/lib/dom/selectorRegistry.ts
+++ b/src/platform-implementation-js/lib/dom/selectorRegistry.ts
@@ -1,0 +1,88 @@
+import {
+  GMAIL_SELECTORS,
+  type SelectorCandidates,
+  type SelectorKey,
+  type SelectorOverrides,
+} from '../../dom-driver/gmail/selectors';
+import mergeSelectorOverrides, {
+  type InvalidSelectorReporter,
+} from './mergeSelectorOverrides';
+import { SelectorError } from './querySelectorOrFail';
+
+export type SelectorRegistryOptions = {
+  /** Additional selector candidates to PREPEND to the SDK bundled defaults. */
+  overrides?: SelectorOverrides;
+  /** Fires for each override candidate dropped as invalid CSS in THIS browser. */
+  onInvalidSelector?: InvalidSelectorReporter;
+};
+
+/**
+ * Install overrides into the bundled defaults, filtering out invalid
+ * selectors. If there are any errors merging the overrides, we fallback
+ * to the SDK bundled defaults.
+ */
+function mergeOrFallBack({
+  overrides,
+  onInvalidSelector,
+}: SelectorRegistryOptions): SelectorCandidates {
+  if (!overrides) {
+    return GMAIL_SELECTORS;
+  }
+
+  try {
+    return mergeSelectorOverrides(
+      GMAIL_SELECTORS,
+      overrides,
+      onInvalidSelector,
+    );
+  } catch {
+    return GMAIL_SELECTORS;
+  }
+}
+
+/**
+ * Resolves named keys to elements, trying each of the key's candidates in
+ * order. Owned by the GmailDriver, so each InboxSDK instance resolves against
+ * its own candidates and a registry is never read before it is built.
+ */
+export default class SelectorRegistry {
+  readonly #candidates: SelectorCandidates;
+
+  constructor(options: SelectorRegistryOptions = {}) {
+    this.#candidates = mergeOrFallBack(options);
+  }
+
+  /**
+   * Resolve a registry key against `el`, returning null on a miss. Use this
+   * method if you want to degrade functionality when a selector is missing,
+   * instead of throwing an error.
+   */
+  querySelectorByKey(
+    el: Document | DocumentFragment | Element,
+    key: SelectorKey,
+  ): HTMLElement | null {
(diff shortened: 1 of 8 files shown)


jfmaggie (approved):
👍 changes lgtm and it makes sense to me to try from a specific example like compose view here as opposed to add the generic registry to all views. 

would def recommend @Macil to take a look. 

one thing it's not super clear to me is the rest of changes, any api changes to allow the config being merged on the consumer side, curious to see that next but not a blocker for this pr

pnavarrc:
> one thing it's not super clear to me is the rest of changes, any api changes to allow the config being merged on the consumer side

Interesting. We can always update later to allow the consumer to apply their own merging logic if we get this feature request or we find a use case.

src/platform-implementation-js/lib/dom/selectorRegistry.ts
```diff
@@ -0,0 +1,71 @@
...
+import mergeSelectorOverrides, {
+  type InvalidSelectorReporter,
+} from './mergeSelectorOverrides';
+import { SelectorError } from './querySelectorOrFail';
+
+/**
+ * Global variable to keep the selector registry, initialized
+ * with the SDK bundled defaults.
+ */
+let candidates: SelectorCandidates = GMAIL_SELECTORS;
```
Macil, line 16: Instead of using module state (which will be shared by all InboxSDK instances), make this be a class. (Not that we have too much of a use-case for multiple InboxSDK instances, but sometimes it is useful for testing, and also the pattern of keeping state out of modules and tied to the InboxSDK/GmailDriver instance is a pattern the InboxSDK follows by default. It also helps make it obvious when something has been properly initialized before using it if you need a reference to a constructed class first.)
Do the logic of loadSelectorRegistry in SelectorRegistry's constructor, and make querySelectorByKey and querySelectorByKeyOrFail be methods on it.

Make there be a `readonly selectors: SelectorRegistry` property on GmailDriver and initialize it in GmailDriver's constructor. (GmailDriver is basically the owner of everything related to an InboxSDK instance.) Make GmailDriver's constructor take an extra argument of a SelectorRegistryOptions object consisting of `overrides?: SelectorOverrides; onInvalidSelector?: InvalidSelectorReporter;` to use when constructing the SelectorRegistry.

This does mean that places like GmailComposeView will need to do stuff like `this.#driver.selectors.querySelectorByKey(...)`, but basically everything has a reference to GmailDriver like this or can easily have it passed in if it is lacking it so this shouldn't be too disruptive to using it.

Everything else in this PR looks good.

pnavarrc: Great suggestions and rationale, that eliminates the timing issue and keep things consistent. Updating!

PR #1312, opened 2026-08-17
feat(selectors): key the compose title bar mutation targets

pnavarrc:
Follow-up to #1308. The registry landed with two keys for the compose title bar, but both resolve elements that are only *found*, never meaningfully mutated.

The changes in this PR allow us to actually override the registry in the parts where the DOM is changed.

- **`setTitleBarColor`** — the keyed `composeView.titleBarTd` only receives a marker class. The `backgroundColor` paint, which is the point of the API, went to three elements found by raw `querySelector`:

  ```
  .nH.Hy.aXJ .pi > .l.o
  .nH.Hy.aXJ .l.m
  .nH.Hy.aXJ .l.m > .l.n
  ```
- **`setTitleBarText`** — `composeView.titleBarTable` is keyed, but the insertion anchor `div.Hp` (whose parent cell the custom `<td>` is placed after, and which gets hidden) was not.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/selectors.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/selectors.ts
@@ -13,10 +13,11 @@
  */
 export const GMAIL_SELECTORS = {
   /**
-   * The compose title-bar table, resolved against the compose root.
+   * The compose title-bar table.
    *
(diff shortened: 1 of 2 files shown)


Macil (approved):

jfmaggie (approved):
lgtm

PR #1316, opened 2026-08-24
fix(message view): restore more menu items in Gmail's current menu

pnavarrc:
Fixes the per-message "More" menu items in Gmail's current menu.

message - more menu items

The problem was caused by two concurrent bugs:

- The SDK couldn't find the ⋮ button using the stored CSS selectors (Gmail changed the markup).
- Gmail build the menu when the menu opens, which raced with the InboxSDK trying to find the items, which caused the InboxSDK to fail intermittently.

The new selectors are identified, but I want to use this selector to prove the remote selector config idea. Once I do, I will add the new selectors into the registry. In the meantime, we can override the selectors (shown in the examples/toolbar-button example).

### Changes

- When creating a menu item, we clone an existing item and change its contents.
- Fix the race condition that was causing the InboxSDK to not find the menu sometimes.
- Update `examples/toolbar-button` with more realistic items, fix the icon size, and override the selectors.

### DIFF
--- a/src/platform-implementation-js/dom-driver/gmail/views/gmail-message-view.ts
+++ b/src/platform-implementation-js/dom-driver/gmail/views/gmail-message-view.ts
@@ -1,6 +1,5 @@
 import sortBy from 'lodash/sortBy';
 import once from 'lodash/once';
-import autoHtml from 'auto-html';
 import * as Kefir from 'kefir';
 import kefirBus from 'kefir-bus';
 import type { Bus } from 'kefir-bus';
@@ -10,6 +9,10 @@ import GmailAttachmentAreaView from './gmail-attachment-area-view';
 import GmailAttachmentCardView from './gmail-attachment-card-view';
 import getUpdatedContact from './gmail-message-view/get-updated-contact';
 import AttachmentIcon from './gmail-message-view/attachment-icon';
+import {
+  createMoreMenuDivider,
+  createMoreMenuItem,
+} from './gmail-message-view/more-menu-item';
 import makeMutationObserverStream from '../../../lib/dom/make-mutation-observer-stream';
 import querySelector from '../../../lib/dom/querySelectorOrFail';
 import makeMutationObserverChunkedStream from '../../../lib/dom/make-mutation-observer-chunked-stream';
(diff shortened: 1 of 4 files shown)


src/platform-implementation-js/dom-driver/gmail/views/gmail-message-view.ts
```diff
@@ -381,6 +391,24 @@ class GmailMessageView {
...
+
+    if (openMoreMenu) {
+      return Kefir.constant(openMoreMenu);
+    }
+
+    return streamWaitFor(
+      () => this.#getOpenMoreMenu(),
+      MORE_MENU_WAIT_TIMEOUT,
+      MORE_MENU_WAIT_STEP,
+    ).flatMapErrors(() => Kefir.constant(null));
```
hwfwalton, line 409: why null out the errors? were they causing typing issues downstream?

src/platform-implementation-js/dom-driver/gmail/views/gmail-message-view.ts
```diff
@@ -72,7 +82,7 @@ class GmailMessageView {
   #eventStream: Bus<MessageViewDriverEvents, unknown> = kefirBus();
   #stopper = kefirStopper();
   #threadViewDriver: GmailThreadView;
-  #moreMenuItemDescriptors: Array<Record<string, any>>;
+  #moreMenuItemDescriptors: Array<MessageViewToolbarButtonDescriptor>;
```
hwfwalton, line 85: good change (and all the downstream spots). I'm surprised this wasn't already typed

hwfwalton (approved):
really clever approach with cloning and cleaning an existing item! should be more resilient to style changes.
The stream changes are all pretty clean and done idiomatically wrt the sdk codebase.
good call to update the example now (with the override).
fingers crossed for remote selector loading!
lgtm 👍

src/platform-implementation-js/dom-driver/gmail/views/gmail-message-view.ts
pnavarrc: We need the `null` to cleanup the state (mimic "menu closed")

PR #1321, opened 2026-09-08
fix(route-view): insert custom sections below the list toolbar

pnavarrc:
Gmail nests the list toolbar inside the row list wrapper (`.bGI.nH`), so custom sections inserted at the top of the row list container render above the toolbar instead of below it.

Two commits:

- the insertion is anchored to the toolbar, resolved through two new registry keys (`routeView.rowListWrapper`, `routeView.listToolbar`), and falls back to the previous placement when the toolbar is not found
- the route view's page parser now takes the driver's registry and resolves the same wrapper key, rather than hardcoding `.bGI.nH`

Typecheck, tests and lint pass. Verified against live Gmail.


hwfwalton (approved):
Good use of the SelectoryRegistry and overall hardening of the toolbar insert codepath.
lgtm 👍

PR #1322, opened 2026-09-11
fix(route-view): anchor custom sections to the displayed row list wrapper

frew:
Custom sections disappeared on Gmail search. Every list route shares one `.inboxsdk__custom_sections`, and Gmail hides or deletes the wrapper of the list you navigate away from.

- place the container on every list route, not only when it is created
- anchor it below the toolbar of the displayed row list wrapper, or at the top of the list container while that toolbar is missing
- keep re-placing it while the route is open, because Gmail can switch wrappers or render the toolbar after a route adds its sections
- skip the move when the container is already in place
- unit tests for the placement helper and for `querySelectorAllByKey` rung selection
- checked in Gmail with a real extension: inbox to search and back, a fresh load on search then inbox, a second search, and a wrapper switch with no route change


pnavarrc (approved):
The original fix placed the sections container below the visible list toolbar, only on creation. 

I added the container on every route, and adds it again when Gmail shows a different list or adds the toolbar.

PR #1324, opened 2026-09-15
fix(route-view): re-place custom sections on the next animation frame

pnavarrc:
Gmail's row hover toolbar flickers on list routes that have a custom section, since 2.2.24 (#1322).

- the sections container observer sees every class change in the list, including Gmail's row hover class
- it re-placed the container inside the mutation callback, and placement reads computed style while Gmail is still handling the hover
- placement now waits for the next animation frame

## Test plan

- [x] Gmail, pointer held on the toolbar edge of an inbox row with sections: 24 hover class toggles in 3 seconds before, 0 after
- [x] Gmail: sections stay below the displayed list toolbar across inbox, search, tracked emails, drafts and sent
- [x] Gmail: real mouse over the toolbar icons, inbox and search


hwfwalton (approved):
wowza, good find.
Tested locally and fixes the originally reported issue for me.
lgtm 👍

PR #1325, opened 2026-09-16
feat(selectors): key the left nav and app menu lookups

pnavarrc:
Moves the element getter's left nav and app menu lookups onto the selector registry, so a published config can fix them without a release.

Better reviewed commit by commit.

### Changes

- Add 7 keys: `leftNav.root`, `leftNav.appMenu`, `leftNav.appMenuOrNavMenuFallback`, `leftNav.gtalkButtons`, `appMenu.container`, `appMenu.burgerButton`, `appMenu.header`.
- Move 10 lookups in `gmail-element-getter.ts` onto them. Each still returns null on a miss.
- Replace the `APP_MENU, NAV_MENU` selector list with `NAV_MENU` alone, which finds the same element, then delete both constants.
- One commit per key.

### Testing

- `yarn test`, `yarn typecheck` and `yarn lint` pass. Each commit typechecks.
- Not yet checked in Gmail.


src/platform-implementation-js/dom-driver/gmail/selectors.ts
```diff
@@ -156,6 +156,50 @@ export const GMAIL_SELECTORS = {
...
+   * The Hangouts buttons at the bottom of the left nav.
+   * Root: `document`.
+   */
+  'leftNav.gtalkButtons': ['.aeN .aj5.J-KU-Jg'],
+
+  /**
+   * The left nav, present with or without the app menu.
+   * Root: `document`.
+   */
+  'leftNav.navMenu': ['.aeN.WR.nH.oy8Mbf[role=navigation]'],
```
Macil: The purpose of this selector was a bit unclear before but now the description is less accurate. I think the following description would be good as the comment:

This selector matches the app menu if it is or will be present, otherwise if we're in a mode of Gmail without the app menu, then this might match the collapsible panel left nav menu instead. We wait for this selector so we can know when the app menu is ready or if it will never be ready.

Also maybe it should be called "leftNav.appMenuOrNavMenuFallback".

src/platform-implementation-js/dom-driver/gmail/selectors.ts
```diff
@@ -156,6 +156,50 @@ export const GMAIL_SELECTORS = {
...
+   * The left nav column.
+   * Root: `document`.
+   */
+  'leftNav.root': ['.aeN'],
+
+  /**
+   * The left nav when Gmail shows the app menu.
+   * Root: `document`.
+   */
+  'leftNav.appMenu': ['.aeN.WR.a6o.anZ.nH.oy8Mbf[role=navigation]'],
```
Macil, line 171: Keep the url from the old comment because it had a nice visual explanation for what the app menu was (https://support.google.com/mail/answer/11555490)

Macil (changes requested):

Macil (approved):

PR #1326, opened 2026-09-17
feat(selectors): key the thread view lookups

pnavarrc:
Selector coverage group 2: 12 keys move out of `gmail-thread-view.ts` into `GMAIL_SELECTORS`, one per commit.

Keys added:

- `threadView.messageList`, `hiddenNoticeMessage`, `hiddenNoticeCount`, `hiddenNoticeSlot`
- `threadView.labelContainer`, `subjectParent`, `messagesContainer`, `subjectToolbar`, `openMessage`
- `threadView.expandAllButton`, `collapseAllButton`
- `threadView.subjectContainer` and `subjectContainerLegacy`, replacing the two version maps in `addNoticeBar` and the loop over them.

**Notes**

- `SelectorError` at `subjectParent` and `messagesContainer` now names the key, not the raw selector
- the `addNoticeBar subjectContainer` log's `version` now carries the key name, not a date string
- `#findSubjectToolbarElement` drops a local that was assigned and immediately returned

Probably best reviewed commit by commit, starting with https://github.com/InboxSDK/InboxSDK/pull/1326/changes/3cc9d794d04bc34efbaceb2fbaedd1898c52cefa


src/platform-implementation-js/dom-driver/gmail/selectors.ts
```diff
@@ -145,6 +145,101 @@ export const GMAIL_SELECTORS = {
    */
   'threadView.subject': ['.ha h2'],
 
+  /**
+   * The list holding the thread's messages, which custom messages are
+   * inserted into.
+   * Root: thread element.
```
ruthhalteman: when we assert root here, does that mean it's the root that we're selecting from in our code when we use this selector? That could change over time and get our of sync no? What's the value of that being recorded here? Is it just for context on where in the UI the selector is?

ruthhalteman (approved):

src/platform-implementation-js/dom-driver/gmail/selectors.ts
pnavarrc: The root annotation is mostly for context. One idea was to formalize it and make it part of the schema, but it doesn't represent well the complexity of it (the root could be the document, but only after the page transition, or the root could be one of many thread items, etc).

```tsx
{
  'selectorFamily.keyName': {
    selectors: ['selectorA', 'selectorB'],
    root: 'selectorFamily.otherKey'
  },
  // ...
}
```

I'm not sure how useful it is yet, but since I was in doubt, I settled on keeping at least a note of it for now.

